PR #3, opened 2025-04-08
Support registry/namespace/repository as model name

doringeman:

PR #5, opened 2025-04-09
Extract endpoints and add main.go for testing

doringeman:
Built on top of https://github.com/docker/model-runner/pull/3.

Terminal 1:
```
go run main.go
```

Terminal 2:
```
$ curl --unix-socket model-runner.sock localhost/models/ai/smollm2
{"id":"sha256:354bf30d0aa3af413d2aa5ae4f23c66d78980072d1e07a5b0d776e9606a2f0b9","tags":["ai/smollm2"],"created":1742816981,"config":{"format":"gguf","quantization":"IQ2_XXS/Q4_K_M","parameters":"361.82 M","architecture":"llama","size":"256.35 MiB"}}
$ curl --unix-socket model-runner.sock localhost/engines/v1/models/ai/smollm2
{"id":"ai/smollm2","object":"model","created":1742816981,"owned_by":"docker"}
$ curl --unix-socket model-runner.sock localhost/engines/llama.cpp/v1/models/ai/smollm2
{"id":"ai/smollm2","object":"model","created":1742816981,"owned_by":"docker"}
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,0 +1,68 @@
+package main
+
+import (
+	"net"
+	"net/http"
+	"os"
+	"os/signal"
+	"path/filepath"
+	"syscall"
+
+	"github.com/docker/model-distribution/pkg/distribution"
+	"github.com/docker/model-runner/pkg/inference/models"
+	"github.com/sirupsen/logrus"
+)
+
+var log = logrus.New()
+
+func main() {
+	stop := make(chan os.Signal, 1)
+	signal.Notify(stop, os.Interrupt, syscall.SIGTERM)
+
+	userHomeDir, err := os.UserHomeDir()
+	if err != nil {
+		log.Fatalf("Failed to get user home directory: %v", err)
+	}
+	distributionClient, err := distribution.NewClient(
+		distribution.WithStoreRootPath(filepath.Join(userHomeDir, ".docker", "models")),
+	)
+	if err != nil {
+		log.Fatalf("Failed to create distribution client: %v", err)
+	}
+
+	modelManager := models.NewManager(log, distributionClient)
+
+	router := http.NewServeMux()
+	for _, route := range modelManager.GetRoutes() {
+		router.Handle(route, modelManager)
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -0,0 +1,68 @@
...
+		router.Handle(route, modelManager)
+	}
+
+	sockName := "model-runner.sock"
+	if err := os.Remove(sockName); err != nil {
+		if !os.IsNotExist(err) {
+			log.Fatalf("Failed to remove existing socket: %v", err)
+		}
+	}
+	defer os.Remove(sockName)
```
xenoscopic: minor nit: move `defer os.Remove(sockName)` to after `ListenUnix` so that we don't remove a socket that is racily created between the first `os.Remove` and `net.ListenUnix` (i.e. we only want to remove our socket).

main.go
```diff
@@ -0,0 +1,69 @@
...
+		if !os.IsNotExist(err) {
+			log.Fatalf("Failed to remove existing socket: %v", err)
+		}
+	}
+	ln, err := net.ListenUnix("unix", &net.UnixAddr{Name: sockName, Net: "unix"})
+	if err != nil {
+		log.Fatalf("Failed to listen on socket: %v", err)
+	}
+	defer os.Remove(sockName)
+	defer ln.Close()
```
xenoscopic: `sever.Close()` will also close the listener implicitly, so this cold probably be skipped.

main.go
```diff
@@ -0,0 +1,69 @@
...
+
+	stop := make(chan os.Signal, 1)
+	signal.Notify(stop, os.Interrupt, syscall.SIGTERM)
```
xenoscopic: I would move these to the start of `main()` so that no signals kill the process in an uncontrolled fashion during the initialization phase (e.g. when we might be acquiring locks or something).

main.go
doringeman: Right, good catch! Thanks!

PR #8, opened 2025-04-10
Initialize model distribution in model-runner

ilopezluna:
Initialize model distribution in model-runner. This allows to remove model-distribution dependency in Pinata.

### DIFF
--- a/pkg/inference/models/manager.go
+++ b/pkg/inference/models/manager.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/sirupsen/logrus"
 	"html"
 	"net/http"
 
@@ -13,6 +12,7 @@ import (
 	"github.com/docker/model-distribution/pkg/types"
 	"github.com/docker/model-runner/pkg/inference"
 	"github.com/docker/model-runner/pkg/logging"
+	"github.com/sirupsen/logrus"
 )
 
 const (


pkg/inference/models/manager.go
```diff
@@ -33,14 +34,38 @@ type Manager struct {
...
+	Transport http.RoundTripper
+	// UserAgent is the user agent to use.
+	UserAgent string
+}
+
 // NewManager creates a new model's manager.
-func NewManager(log logging.Logger, client *distribution.Client) *Manager {
+func NewManager(log logging.Logger, c ClientConfig) *Manager {
+	// Create the model distribution client.
+	distClient, err := distribution.NewClient(
```
doringeman: ?
```suggestion
	distributionClient, err := distribution.NewClient(
```

pkg/inference/models/manager.go
```diff
@@ -33,14 +34,38 @@ type Manager struct {
...
+		// respond to requests, but may return errors if the client is required.
+	}
+
 	// Create the manager.
 	m := &Manager{
 		log:                log,
 		pullTokens:         make(chan struct{}, maximumConcurrentModelPulls),
 		router:             http.NewServeMux(),
-		distributionClient: client,
+		distributionClient: distClient,
```
doringeman:
```suggestion
		distributionClient: distributionClient,
```

pkg/inference/models/manager.go
```diff
@@ -33,14 +34,38 @@ type Manager struct {
...
 // NewManager creates a new model's manager.
-func NewManager(log logging.Logger, client *distribution.Client) *Manager {
+func NewManager(log logging.Logger, c ClientConfig) *Manager {
+	// Create the model distribution client.
+	distributionClient, err := distribution.NewClient(
+		distribution.WithStoreRootPath(c.StoreRootPath),
+		distribution.WithLogger(c.Logger),
+		distribution.WithTransport(c.Transport),
+		distribution.WithUserAgent(c.UserAgent),
+	)
```
doringeman, line 56: If `c` doesn't has `c.Transport` set, `c.Transport` is `nil` because it's an interface.
And we don't want to set it to `nil` as it'll cause a `panic` on `pull`.
E.g.,
```
curl --unix-socket model-runner.sock localhost/models/create -X POST -d '{"from": "ai/smollm2"}'
```

I think a possible approach to work around this is:
```
diff --git a/pkg/inference/models/manager.go b/pkg/inference/models/manager.go
index 577f588..da16dbc 100644
--- a/pkg/inference/models/manager.go
+++ b/pkg/inference/models/manager.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/sirupsen/logrus"
 	"html"
 	"net/http"
 
@@ -13,6 +12,7 @@ import (
 	"github.com/docker/model-distribution/pkg/types"
 	"github.com/docker/model-runner/pkg/inference"
 	"github.com/docker/model-runner/pkg/logging"
+	"github.com/sirupsen/logrus"
 )
 
 const (
@@ -48,12 +48,15 @@ type ClientConfig struct {
 // NewManager creates a new model's manager.
 func NewManager(log logging.Logger, c ClientConfig) *Manager {
 	// Create the model distribution client.
-	distributionClient, err := distribution.NewClient(
+	opts := []distribution.Option{
 		distribution.WithStoreRootPath(c.StoreRootPath),
 		distribution.WithLogger(c.Logger),
-		distribution.WithTransport(c.Transport),
 		distribution.WithUserAgent(c.UserAgent),
-	)
+	}
+	if c.Transport != nil {
+		opts = append(opts, distribution.WithTransport(c.Transport))
+	}
+	distributionClient, err := distribution.NewClient(opts...)
 	if err != nil {
 		log.Errorf("Failed to create distribution client: %v", err)
 		// Continue without distribution client. The model manager will still
```

doringeman: Same for all "options" actually...

pkg/inference/models/manager.go
```diff
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/sirupsen/logrus"
```
doringeman: Let's sort the imports to be consistent with the current coding style, even if it's not automatically verified, yet. 😓 
```
diff --git a/pkg/inference/models/manager.go b/pkg/inference/models/manager.go
index 577f588..8e0ae2b 100644
--- a/pkg/inference/models/manager.go
+++ b/pkg/inference/models/manager.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/sirupsen/logrus"
 	"html"
 	"net/http"
 
@@ -13,6 +12,7 @@ import (
 	"github.com/docker/model-distribution/pkg/types"
 	"github.com/docker/model-runner/pkg/inference"
 	"github.com/docker/model-runner/pkg/logging"
+	"github.com/sirupsen/logrus"
 )
 
 const (
```

PR #11, opened 2025-04-16
Add endpoint for tagging a model

ekcasey:
Adds an endpoint for tagging models that is analogous to the docker engine [endpoint](https://docs.docker.com/reference/api/engine/version/v1.45/#tag/Image/operation/ImageTag) for tagging images.

Method: `POST`
Path: `/models/{name}/tag`
Query Params:
  - `repo` (required)
  - `tag` (optional: defaults to `latest`)
  
 Can be used by the CLI to implement `docker model tag`.

### DIFF
--- a/pkg/inference/models/manager.go
+++ b/pkg/inference/models/manager.go
@@ -92,6 +92,7 @@ func (m *Manager) routeHandlers() map[string]http.HandlerFunc {
 		"GET " + inference.ModelsPrefix:                                       m.handleGetModels,
 		"GET " + inference.ModelsPrefix + "/{name...}":                        m.handleGetModel,
 		"DELETE " + inference.ModelsPrefix + "/{name...}":                     m.handleDeleteModel,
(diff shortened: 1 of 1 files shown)


pkg/inference/models/manager.go
```diff
@@ -288,6 +289,53 @@ func (m *Manager) handleOpenAIGetModel(w http.ResponseWriter, r *http.Request) {
...
+			http.Error(w, err.Error(), http.StatusNotFound)
+			return
+		}
+
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	// Respond with success.
+	w.WriteHeader(http.StatusOK)
```
xenoscopic:
```suggestion
	w.WriteHeader(http.StatusCreated)
```

xenoscopic (approved):
LGTM!  Just had one small suggestion.

PR #19, opened 2025-04-23
Enable basic windows/arm64 support

p1-0tr:


pkg/inference/backends/llamacpp/llamacpp.go
```diff
@@ -74,7 +75,8 @@ func (l *llamaCpp) Install(ctx context.Context, httpClient *http.Client) error {
 	// never support it on Intel Macs.
 	if runtime.GOOS == "linux" {
 		return errors.New("not implemented")
-	} else if (runtime.GOOS == "darwin" && runtime.GOARCH == "amd64") || (runtime.GOOS == "windows" && runtime.GOARCH == "arm64") {
+	} else if (runtime.GOOS == "darwin" && runtime.GOARCH == "amd64") ||
+		(runtime.GOOS == "windows" && !slices.Contains([]string{"amd64", "arm64"}, runtime.GOARCH)) {
```
xenoscopic: Static checks for `runtime.GOARCH == "amd64" || runtime.GOARCH == "arm64"` might be slightly more efficient since they'll get constant folded.

doringeman (approved):

xenoscopic (approved):

PR #20, opened 2025-04-23
Adds makefile

ilopezluna:
And updates the README file with examples of how to use it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 .idea/
+model-runner
+model-runner.sock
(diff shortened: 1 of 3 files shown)


xenoscopic (approved):
LGTM.  If this is just in service of the `Dockerfile`, then I think it's good to go.  If we want it to run in Windows environments (e.g. Mingw-w64) then maybe handling the `.exe` suffix on the `-o $(APP_NAME)` flag in `go build` would be worthwhile, but frankly I don't think we'll be building this much on Windows (and I'm fine just running a manual `go build .` there). Getting a GNU-compatible toolchain on Windows is a pain.

PR #22, opened 2025-04-24
Dockerize

ilopezluna:
# Add Docker Support and TCP Port Access

This PR adds Docker support and TCP port access to the model-runner service.

## Changes

- Added Dockerfile with multi-stage build
- Added TCP port support (default: 8080) alongside Unix socket
- Added Docker-related Makefile targets (`docker-build`, `docker-run`)
- Updated documentation with Docker usage and API examples
- Added environment variable configuration for port and llama.cpp path
- Updated dependencies

## Usage

```bash
# Run in Docker
make docker-run

# Customize port and model storage
make docker-run PORT=3000 MODELS_PATH=/path/to/models
```

The service now supports both Unix socket and TCP port access, with persistent model storage through Docker volumes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 .idea/
+model-runner
+model-runner.sock
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,3 @@
 .idea/
 model-runner
 model-runner.sock
-model-runner
-model-runner.sock
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 .idea/
 model-runner
 model-runner.sock
+model-runner
+model-runner.sock
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,13 @@
 # syntax=docker/dockerfile:1
 
-ARG GO_VERSION=1.23.7
+ARG GO_VERSION=1.24.2
 ARG LLAMA_SERVER_VERSION=latest
 ARG LLAMA_BINARY_PATH=/com.docker.llama-server.native.linux.cpu.amd64
 
-FROM golang:${GO_VERSION}-alpine AS builder
+FROM golang:${GO_VERSION}-bookworm AS builder
 
 # Install git for go mod download if needed
-RUN apk add --no-cache git
+RUN apt-get update && apt-get install -y --no-install-recommends git && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /app
 
@@ -31,7 +31,7 @@ RUN --mount=type=cache,target=/go/pkg/mod \
 FROM docker/docker-model-backend-llamacpp:${LLAMA_SERVER_VERSION} AS llama-server
 
 # --- Final image ---
-FROM debian:bookworm-slim AS final
+FROM ubuntu:24.04 AS final
 
 # Create non-root user
 RUN groupadd --system modelrunner && useradd --system --gid modelrunner modelrunner
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,62 @@
+# syntax=docker/dockerfile:1
+
+ARG GO_VERSION=1.23.7
+ARG LLAMA_SERVER_VERSION=latest
+ARG LLAMA_BINARY_PATH=/com.docker.llama-server.native.linux.cpu.amd64
+
+FROM golang:${GO_VERSION}-alpine AS builder
+
+# Install git for go mod download if needed
+RUN apk add --no-cache git
+
+WORKDIR /app
+
+# Copy go mod/sum first for better caching
+COPY --link go.mod go.sum ./
+
+# Download dependencies (with cache mounts)
+RUN --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=cache,target=/root/.cache/go-build \
+    go mod download
+
+# Copy the rest of the source code
+COPY --link . .
+
+# Build the Go binary (static build)
+RUN --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=cache,target=/root/.cache/go-build \
+    CGO_ENABLED=0 GOOS=linux go build -ldflags="-s -w" -o model-runner ./main.go
+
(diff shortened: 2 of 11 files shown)


.gitignore
```diff
@@ -1 +1,5 @@
 .idea/
+model-runner
+model-runner.sock
+model-runner
+model-runner.sock
```
kiview: Duplicate?

Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+# Build the Go binary (static build)
+RUN --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=cache,target=/root/.cache/go-build \
+    CGO_ENABLED=0 GOOS=linux go build -ldflags="-s -w" -o model-runner ./main.go
+
+# --- Get llama.cpp binary ---
+FROM docker/docker-model-backend-llamacpp:${LLAMA_SERVER_VERSION} AS llama-server
+
+# --- Final image ---
+FROM alpine:latest AS final
```
kiview: **thought**: I would be initial concerned about using Alpine as the base because of musl, so maybe Debian (or another base image providing glibc) would be a safer decision.

Makefile
```diff
@@ -0,0 +1,69 @@
...
+docker-run-socket: docker-build clean
+	mkdir -p $(SOCKET_PATH)
+	chmod 777 $(SOCKET_PATH)
+	docker run --rm \
+		-v "$(SOCKET_PATH):/var/run/model-runner" \
+		-e MODEL_RUNNER_SOCK=/var/run/model-runner/model-runner.sock \
+		$(DOCKER_IMAGE)
+	@echo ""
+	@echo "Socket available at: $(SOCKET_PATH)/model-runner.sock"
+	@echo "Example usage: curl --unix-socket $(SOCKET_PATH)/model-runner.sock http://localhost/models"
```
kiview: **note**: Echo commands won't be shown to user, since `docker run` runs without daemon flag and is hence blocking.

Makefile
```diff
@@ -0,0 +1,69 @@
...
+
+# Run in Docker container
+docker-run: docker-build
+	docker run --rm $(DOCKER_IMAGE)
+
+# Run in Docker container with socket accessible from host
+docker-run-socket: docker-build clean
+	mkdir -p $(SOCKET_PATH)
+	chmod 777 $(SOCKET_PATH)
+	docker run --rm \
```
kiview, line 52:
```suggestion
	docker run --rm -d \
```

.gitignore
ilopezluna: No, are different things, `model-runner` is the binary of the app and `model-runner.sock` is the unix socket

Dockerfile
ilopezluna: @kiview would `debian:bookworm-slim` be a good candidate?

.gitignore
xenoscopic: I think he's referring to the existing surrounding lines in the `.gitignore`.  It looks like `model-runner` and `model-runner.sock` are already in that file.

Dockerfile
xenoscopic: I would use `ubuntu:24.04`.  NVIDIA seems to use Ubuntu in all of their CDI docs and produces Ubuntu tags for `nvidia/cuda` but not Debian, so I'd say Ubuntu is probably the better compatibility bet.  We'll also get a newer compiler toolchain.  We're not really concerned with glibc or libstdc++ symbol versioning inside the image, we just need to make sure whatever toolchain we use to build the llama.cpp backend is the same version (or newer) than what's in the model-runner image.

Dockerfile
```diff
@@ -0,0 +1,63 @@
+# syntax=docker/dockerfile:1
+
+ARG GO_VERSION=1.23.7
+ARG LLAMA_SERVER_VERSION=latest
+ARG LLAMA_BINARY_PATH=/com.docker.llama-server.native.linux.cpu.amd64
+
+FROM golang:${GO_VERSION}-alpine AS builder
```
xenoscopic: I'd also aim to use a glibc-compatbile Go image for the compilation too (even with `CGO_ENABLED=false`).  If we use `ubuntu:24.04` for the production images, then `golang:1.24.2-bookworm` tag should be fine (Debian Bookworm ships with glibc 2.36 / GCC 12.2.0, Ubuntu 24.04 ships with glibc 2.39 / GCC 13.2).

Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+# Create non-root user
+RUN groupadd --system modelrunner && useradd --system --gid modelrunner modelrunner
+
+# Install ca-certificates for HTTPS
+RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
+
+# Create directories for the socket file and llama.cpp binary, and set proper permissions
+RUN mkdir -p /var/run/model-runner /app/bin /home/modelrunner/.docker/models && \
```
xenoscopic: It might be more idiomatic to store model data in `/var/lib/model-runner/models`, or something top-level like `/models`.

Makefile
```diff
@@ -20,11 +28,42 @@ run: build
...
 	rm -f model-runner.sock
+	rm -rf $(MODELS_PATH)
+
+# Run tests
+test:
+	go test -v ./...
+
+# Build Docker image
+docker-build:
+	docker build --platform linux/amd64 \
```
doringeman, line 39: `--platform linux/amd64` because there's not a llama.cpp build for linux/arm64 yet, right?

ilopezluna: @doringeman yes

xenoscopic (approved):
LGTM!

Makefile
xenoscopic: Once this gets merged, I'll add a `linux` / `arm64` build for the llama.cpp backend and then add a multi-arch image build.  I want to play with this on the Raspberry Pi.

doringeman (approved):

PR #30, opened 2025-05-07
Add NormalizedServeMux

doringeman:
Add `NormalizedServeMux` implementation that wraps the standard http.ServeMux with automatic path normalization.

The new implementation:
- Automatically normalizes and prevents incorrect route handling for paths with consecutive slashes (e.g., "//models" → "/models").
- Uses Go's `path.Clean()` for consistent path normalization.
- Maintains the same interface as `http.ServeMux` for easy drop-in replacement.

Without this patch, using `DMR_HOST=http://localhost:8080/` with a trailing `/` will make `docker model pull` fail.
```
$ make docker-run
$ DMR_HOST=http://localhost:8080/ docker model pull ai/smollm2
```

### DIFF
--- a/pkg/routing/routing.go
+++ b/pkg/routing/routing.go
@@ -1,0 +1,24 @@
+package routing
+
+import (
+	"net/http"
+	"path"
+	"strings"
+)
+
+type NormalizedServeMux struct {
+	*http.ServeMux
+}
+
+func NewNormalizedServeMux() *NormalizedServeMux {
+	return &NormalizedServeMux{http.NewServeMux()}
+}
+
+func (nm *NormalizedServeMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if strings.Contains(r.URL.Path, "//") {
+		normalizedPath := path.Clean(r.URL.Path)
+		r.URL.Path = normalizedPath
+	}
+
+	nm.ServeMux.ServeHTTP(w, r)
+}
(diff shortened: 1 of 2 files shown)


pkg/routing/routing.go
```diff
@@ -0,0 +1,50 @@
...
+	"path"
+	"strings"
+)
+
+type NormalizedServeMux interface {
+	http.Handler
+	// Handle is a method implemented by http.ServeMux.
+	Handle(pattern string, handler http.Handler)
+	// HandleFunc is a method implemented by http.ServeMux.
+	HandleFunc(pattern string, handler func(http.ResponseWriter, *http.Request))
```
xenoscopic: Could potentially just do

```
type NormalizedServeMux struct {
    *http.ServeMux
}

func (nm *NormalizedServeMux) ServeHTTP(...) {
    ...
}
```

doringeman: And how would it allow registering handles?

xenoscopic: The embedding would expose all of the public methods on `ServeMux`, so you'd get all the `Handle*` methods exposed for free, and the `ServeHTTP` method on the outer method would take priority.

doringeman: Right, I misread http.ServeMux as http.Handler.
I think so, let me try.

doringeman: Force-pushed. Nice one, thank you!

xenoscopic (approved):
Cool, thanks for adding this fix!  ServeMux is so weird...

PR #33, opened 2025-05-08
Remove unused env vars from Makefile and update README

doringeman:
Remove unused TARGET_OS and ACCEL.
Update README to mention LLAMA_SERVER_VERSION and LLAMA_SERVER_VARIANT.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,6 @@ GO_VERSION := 1.23.7
 LLAMA_SERVER_VERSION := v0.0.4
 LLAMA_SERVER_VARIANT := cpu
 BASE_IMAGE := ubuntu:24.04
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -4,8 +4,6 @@ GO_VERSION := 1.23.7
 LLAMA_SERVER_VERSION := v0.0.4
```
p1-0tr: nit: should be `latest`?

doringeman: Good point, thanks!
Fixed in https://github.com/docker/model-runner/pull/33/commits/f826ff1830b0830de0b9fb6b93eb55d89c13e0a0.

p1-0tr (approved):

PR #35, opened 2025-05-09
llamacpp: Use --host instead of DD_INF_UDS

p1-0tr:


ilopezluna (approved):

xenoscopic (approved):
LGTM!

Did llama.cpp's server start supporting UDS natively?

p1-0tr:
> LGTM!
> 
> Did llama.cpp's server start supporting UDS natively?

Yup, I contributed the patch a while back.

PR #40, opened 2025-05-14
Add pull model tests

ilopezluna:
This PR:
- Changes the default folder of model store to `models-store`
- Add `models-store` to `.gitignore`

Otherwise the files in `models` package are ignored by git.

And add a test to verify expected content type of the progress message during model pull.
For that it creates a test registry, builds a model using `dummy.gguf` and then it pulls it


xenoscopic (approved):
Looks great - thanks!

doringeman (approved):
Nice!

PR #41, opened 2025-05-19
configure backend args

ilopezluna:
@doringeman @xenoscopic maybe you have already discussed a different approach to customize `llama.cpp` args, if so we can close this PR. If not I think this could be a good option because its easy to extend it to other backends by implementing the `BackendConfig` interface. 

Note: I'm using `-ngl` to always store the layers in VRAM, in our prev implementation `win/arm64` was only setting this value on Q4 quantization, let me know if you prefer to keep the same behavior
cc @p1-0tr


main.go
```diff
@@ -129,3 +135,48 @@ func main() {
...
+
+	// If no environment variables are set, use default configuration
+	if argsStr == "" {
+		return nil // nil will cause the backend to use its default configuration
+	}
+
+	// Split the string by spaces, respecting quoted arguments
+	args := splitArgs(argsStr)
+	log.Infof("Using custom arguments: %v", args)
+	return &llamacpp.Config{
```
p1-0tr, line 168: nit: Perhaps we should error out if `LLAMA_ARGS` tries to set args we need to control (`--model`, `--host`, `--embeddings`, and `--mmproj` in the future)

p1-0tr (approved):
One small nit, but overall LGTM :)

doringeman (approved):

PR #48, opened 2025-05-22
llamacpp: Skip model warmup

p1-0tr:


ericcurtin:
This could be useful...

The warmup delays initialisation (and you can just start doing real inferencing to warm up).

It seems the warmup is primarily just there so llama.cpp isn't behind on benchmarks (vLLM does this too).

But ultimately is does slow you down at the end of the day as you have to wait for warmup to finish.

I've also seen warmup get stuck on AMD platforms:

https://github.com/containers/ramalama/pull/1477

ericcurtin:
Closing, it's a one arg commit anyway, just as easy open a new pr if we want it

PR #56, opened 2025-05-26
Allow unloading multiple models at once

p1-0tr:
Once we enable running multiple models at once, it will be useful to be able to unload multiple at a time. In preparation for that, make the unload request accept multiple model tags, and evict those models.

### DIFF
--- a/pkg/inference/scheduling/loader.go
+++ b/pkg/inference/scheduling/loader.go
@@ -209,7 +209,10 @@ func (l *loader) Unload(ctx context.Context, unload UnloadRequest) int {
 		if unload.All {
 			return l.evict(false)
 		} else {
-			return l.evictRunner(unload.Backend, unload.Model)
+			for _, model := range unload.Models {
+				l.evictRunner(unload.Backend, model)
+			}
+			return len(l.runners)
 		}
 	}()
 }
--- a/pkg/inference/scheduling/api.go
+++ b/pkg/inference/scheduling/api.go
@@ -64,9 +64,9 @@ type DiskUsage struct {
 
 // UnloadRequest is used to specify which models to unload.
 type UnloadRequest struct {
-	All     bool   `json:"all"`
-	Backend string `json:"backend"`
-	Model   string `json:"model"`
+	All     bool     `json:"all"`
+	Backend string   `json:"backend"`
+	Models  []string `json:"models"`
 }
 
 // UnloadResponse is used to return the number of unloaded runners (backend, model).


pkg/inference/scheduling/loader.go
```diff
@@ -209,7 +209,11 @@ func (l *loader) Unload(ctx context.Context, unload UnloadRequest) int {
 		if unload.All {
 			return l.evict(false)
 		} else {
-			return l.evictRunner(unload.Backend, unload.Model)
+			evicted := 0
+			for _, model := range unload.Models {
+				evicted += l.evictRunner(unload.Backend, model)
```
xenoscopic: The result of `evictRunner` is the number of models remaining after eviction, not the number evicted.  And in any case, we want the anonymous `func() int` to return the number remaining after eviction.

Edit: Ignore the suggestion in my original comment, see below.

xenoscopic (changes requested):

pkg/inference/scheduling/loader.go
xenoscopic: I think the better thing to do would actually just be to not compute `evicted` here and just return `len(l.runners)` in the `else` case of the anonymous `func() int`.

pkg/inference/scheduling/loader.go
```diff
@@ -209,7 +209,11 @@ func (l *loader) Unload(ctx context.Context, unload UnloadRequest) int {
...
+			}
+			return evicted
```
xenoscopic:
```suggestion
			for _, model := range unload.Models {
				l.evictRunner(unload.Backend, model)
			}
			return len(l.runners)
```

pkg/inference/scheduling/loader.go
p1-0tr: I'm not sure if returning the number of remaining runners is useful. Somehow evicted count seems more informative. BTW can an `unload` be handled in parallel with another request, or do we serialise them?

xenoscopic: The `Unload` method does return the number evicted (there's a `len(l.runners) - func() int` computation happening).

pkg/inference/scheduling/loader.go
```diff
@@ -198,7 +198,7 @@ func (l *loader) evictRunner(backend, model string) int {
 	return len(l.runners)
 }
 
-// Unload unloads runners and returns the number of unloaded runners.
+// Unload unloads runners and returns the number of remaining runners.
```
xenoscopic:
```suggestion
// Unload unloads runners and returns the number of unloaded runners.
```
You can leave this one as `unloaded`; the method is returning `len(l.runners) - # remaining` (i.e. number unloaded).

pkg/inference/scheduling/loader.go
p1-0tr: 🤦 true, I don't know how I managed to miss that

PR #58, opened 2025-05-28
Add metrics tracker for model usage

doringeman:
Run ⬇️ in a terminal and look for `component=metrics` logs.
```
DEBUG=1 make docker-run
```

To trigger some logs, in a second terminal you can run:
```
MODEL_RUNNER_HOST=http://localhost:8080 docker model run ai/smollm2 'hi'
```

`func (t *Tracker) TrackModel` is non-blocking.
If we want to make the `doNotTrack` check slightly faster, we could expose this struct member and do a function call only if it's false, hence avoiding a jump.

### DIFF
--- a/pkg/inference/scheduling/scheduler.go
+++ b/pkg/inference/scheduling/scheduler.go
@@ -48,6 +48,7 @@ func NewScheduler(
 	modelManager *models.Manager,
 	httpClient *http.Client,
 	allowedOrigins []string,
+	tracker *metrics.Tracker,
 ) *Scheduler {
 	// Create the scheduler.
 	s := &Scheduler{
@@ -58,7 +59,7 @@ func NewScheduler(
 		installer:      newInstaller(log, backends, httpClient),
 		loader:         newLoader(log, backends, modelManager),
 		router:         http.NewServeMux(),
-		tracker:        metrics.NewTracker(httpClient, log.WithField("component", "metrics")),
+		tracker:        tracker,
 	}
 
 	// Register routes.
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -18,6 +18,7 @@ type Tracker struct {
 	doNotTrack bool
 	transport  http.RoundTripper
 	log        logging.Logger
+	userAgent  string
 }
 
 type TrackerRoundTripper struct {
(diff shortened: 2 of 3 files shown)


pkg/metrics/metrics.go
```diff
@@ -0,0 +1,74 @@
...
+		return
+	}
+
+	go t.trackModel(model)
+}
+
+func (t *Tracker) trackModel(model types.Model) {
+	id, err := model.ID()
+	if err != nil {
+		t.log.Errorf("failed to get model ID for %v: %v", model.Tags(), err)
```
xenoscopic: I think this path is missing a `return`

xenoscopic (approved):
LGTM!

doringeman:
I won't merge this yet as I'm planning to use the registry API instead.

p1-0tr (approved):

pkg/inference/scheduling/scheduler.go
```diff
@@ -194,14 +198,17 @@ func (s *Scheduler) handleOpenAIInference(w http.ResponseWriter, r *http.Request
...
+		if err != nil {
 			if errors.Is(err, distribution.ErrModelNotFound) {
 				http.Error(w, err.Error(), http.StatusNotFound)
 			} else {
 				http.Error(w, "model unavailable", http.StatusInternalServerError)
 			}
 			return
 		}
+		// Non-blocking call to track the model usage.
+		s.tracker.TrackModel(model)
```
xenoscopic, line 212: Maybe a stupid question, but should we disable this in Docker Desktop?  Otherwise I'm not sure how we would delineate Docker CE statistics.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,85 @@
...
+	ref, err := name.ParseReference(tags[0])
+	if err != nil {
+		t.log.Errorf("Error parsing reference: %v\n", err)
+		return
+	}
+	if _, err = remote.Head(ref, []remote.Option{
+		remote.WithAuthFromKeychain(authn.DefaultKeychain),
+		remote.WithTransport(t.transport),
+		remote.WithUserAgent("docker-model-runner"),
+	}...); err != nil {
```
xenoscopic:
```suggestion
	if _, err = remote.Head(ref,
		remote.WithAuthFromKeychain(authn.DefaultKeychain),
		remote.WithTransport(t.transport),
		remote.WithUserAgent("docker-model-runner"),
	); err != nil {
```
Can't these just be passed in as args rather than structured / destructured?

pkg/inference/scheduling/scheduler.go
doringeman: Let's set another User-Agent for Docker Desktop. 😁 
Done in https://github.com/docker/model-runner/pull/58/commits/d24babea539d23e9cc2f093f05292bbe0d55ef9a, as well as a way to disable tracking programmatically.

xenoscopic (approved):
Looks great, thanks!

PR #59, opened 2025-05-28
Return error from the inference backend while waiting

p1-0tr:
If the inference backend errors out while we wait for it to become ready we should not wait for the timeout to expire, but rather surface the error to the user. So change the runner.wait routine to check for the inference backend termination and return an error in case it terminates unexpectedly.

### DIFF
--- a/pkg/inference/scheduling/runner.go
+++ b/pkg/inference/scheduling/runner.go
@@ -29,6 +29,10 @@ const (
 // long to initialize and respond to a readiness request.
 var errBackendNotReadyInTime = errors.New("inference backend took too long to initialize")
 
+// errBackendQuitUnexpectedly indicates that an inference backend terminated
+// unexpectedly
+var errBackendQuitUnexpectedly = errors.New("inference backend quit unexpectedly")
+
 // RunnerSocketPath determines the Unix domain socket path used to communicate
 // with runners at the specified slot. It can be overridden during init().
 var RunnerSocketPath = func(slot int) (string, error) {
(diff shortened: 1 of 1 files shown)


xenoscopic (approved):
LGTM and definitely a better UX.

What would be really helpful (if it's an easy addition) would be modifying `llamaCpp.Run` to capture some portion of `llamaCppProcess.Stderr` in-memory and (if non-empty / valid UTF-8) use that to craft an error message (since the result from `llamaCppProcess.Wait()` probably won't be very informative).  I've done something like that [here](https://github.com/mutagen-io/mutagen/blob/03c982a3a34e1dfe496f4bcd4ee7ced34307db68/pkg/agent/dial.go#L84-L103) and it's been very helpful in debugging.

(In this example you could probably skip the `NewValveWriter` part because `Run` doesn't return until the process has terminated. And feel free to borrow anything from `github.com/mutagen-io/mutagen/pkg/stream` if you want; Docker owns the copyright now).

p1-0tr:
> LGTM and definitely a better UX.
> 
> What would be really helpful (if it's an easy addition) would be modifying `llamaCpp.Run` to capture some portion of `llamaCppProcess.Stderr` in-memory and (if non-empty / valid UTF-8) use that to craft an error message (since the result from `llamaCppProcess.Wait()` probably won't be very informative). I've done something like that [here](https://github.com/mutagen-io/mutagen/blob/03c982a3a34e1dfe496f4bcd4ee7ced34307db68/pkg/agent/dial.go#L84-L103) and it's been very helpful in debugging.
> 
> (In this example you could probably skip the `NewValveWriter` part because `Run` doesn't return until the process has terminated. And feel free to borrow anything from `github.com/mutagen-io/mutagen/pkg/stream` if you want; Docker owns the copyright now).

Made a new PR for that - https://github.com/docker/model-runner/pull/62 . Still needs a little polish (the tail buffer is implemented very naively), though.

PR #60, opened 2025-05-28
Allow disabling inference server auto update

p1-0tr:
Being able to disable the inference server auto update process is useful for testing. Eventually we should also provide it as an option to DD users. So, add an option to disable the inference server auto update.

### DIFF
--- a/pkg/inference/backends/llamacpp/download.go
+++ b/pkg/inference/backends/llamacpp/download.go
@@ -27,12 +27,22 @@ const (
 var (
 	ShouldUseGPUVariant     bool
 	ShouldUseGPUVariantLock sync.Mutex
+	ShouldUpdateServer      = true
+	ShouldUpdateServerLock  sync.Mutex
 	errLlamaCppUpToDate     = errors.New("bundled llama.cpp version is up to date, no need to update")
 )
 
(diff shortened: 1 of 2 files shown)


doringeman (approved):
LGTM!
Maybe it's worth adding the env var to the Makefile as well?

pkg/inference/backends/llamacpp/download.go
```diff
@@ -27,12 +27,22 @@ const (
 var (
 	ShouldUseGPUVariant     bool
 	ShouldUseGPUVariantLock sync.Mutex
+	ShouldUpdateServer      = true
+	ShouldUpdateServerLock  sync.Mutex
 	errLlamaCppUpToDate     = errors.New("bundled llama.cpp version is up to date, no need to update")
```
xenoscopic, line 32: Not really related to this PR, but maybe we should replace `errLlamaCppUpToDate` with a `return nil` as well and a log entry (like you've done with the `update disabled` case).  I worry it's sort of confusing in the status output.

xenoscopic (approved):
LGTM

pkg/inference/backends/llamacpp/download.go
p1-0tr: I hit merge, and quickly realised why we needed to return an error in both cases - https://github.com/docker/model-runner/pull/63 . I'll have a look at reworking that code a bit so that we don't need to return those.

PR #62, opened 2025-05-29
Include llama-server output in server error

p1-0tr:
The llama-server exit status won't be sufficiently informative in most
cases. So gather llama-server's stderr in a tail buffer and include it
in the error returned from llamacpp.Run.

### DIFF
--- a/pkg/inference/backends/llamacpp/llamacpp.go
+++ b/pkg/inference/backends/llamacpp/llamacpp.go
@@ -4,18 +4,21 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"io/fs"
 	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strings"
 
 	"github.com/docker/model-runner/pkg/diskusage"
 	"github.com/docker/model-runner/pkg/inference"
 	"github.com/docker/model-runner/pkg/inference/config"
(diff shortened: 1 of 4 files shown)


pkg/inference/backends/llamacpp/llamacpp.go
```diff
@@ -162,9 +165,11 @@ func (l *llamaCpp) Run(ctx context.Context, socket, model string, mode inference
 		}
 		return llamaCppProcess.Process.Signal(os.Interrupt)
 	}
+	tailBuf := tailbuffer.NewTailBuffer(1024)
 	serverLogStream := l.serverLog.Writer()
-	llamaCppProcess.Stdout = serverLogStream
-	llamaCppProcess.Stderr = serverLogStream
+	out := io.MultiWriter(serverLogStream, tailBuf)
+	llamaCppProcess.Stdout = out
```
xenoscopic: I would keep `Stdout` pointing at `serverLogStream`; I think it will produce too much noise to extract a sensible error.  I would only tee `Stderr`.  It will also prevent output from interleaving in the buffer.

xenoscopic: Of course, that's dependent on how llama.cpp outputs its errors, but I think it delineates them into stderr.

p1-0tr: Done :)

xenoscopic (approved):
LGTM!

PR #68, opened 2025-06-05
Reload defunct runners

p1-0tr:
In case a runner becomes defunct, e.g. as a result of a backend crash it would be neat to be able to reload it. So, if the loader finds runner, have it check if the runner is still alive, and create a new one if the runner is defunct.

### DIFF
--- a/pkg/inference/scheduling/loader.go
+++ b/pkg/inference/scheduling/loader.go
@@ -154,15 +154,22 @@ func (l *loader) broadcast() {
 }
 
 // evict evicts all unused runners from the loader. If idleOnly is true, then
-// only those unused runners which are considered "idle" (based on usage
-// timestamp) are evicted. The caller must hold the loader lock. It returns the
-// number of remaining runners.
+// only those unused, but functioning, runners which are considered "idle" (based
+// on usage timestamp) are evicted. Defunct (e.g. crashed) runners will be evicted
+// regardless of whether they are considered "idle". The caller must hold the loader
+// lock. It returns the number of remaining runners.
 func (l *loader) evict(idleOnly bool) int {
 	now := time.Now()
 	for r, slot := range l.runners {
 		unused := l.references[slot] == 0
 		idle := unused && now.Sub(l.timestamps[slot]) > runnerIdleTimeout
-		if unused && (!idleOnly || idle) {
+		defunct := false
+		select {
+		case <-l.slots[slot].done:
+			defunct = true
+		default:
+		}
+		if unused && (!idleOnly || idle || defunct) {
 			l.log.Infof("Evicting %s backend runner with model %s in %s mode",
 				r.backend, r.model, r.mode,
 			)
@@ -179,11 +186,11 @@ func (l *loader) evict(idleOnly bool) int {
 
 // evictRunner evicts a specific runner. The caller must hold the loader lock.
 // It returns the number of remaining runners.
-func (l *loader) evictRunner(backend, model string) int {
+func (l *loader) evictRunner(backend, model string, mode inference.BackendMode) int {
 	allBackends := backend == ""
 	for r, slot := range l.runners {
 		unused := l.references[slot] == 0
-		if unused && (allBackends || r.backend == backend) && r.model == model {
+		if unused && (allBackends || r.backend == backend) && r.model == model && r.mode == mode {
 			l.log.Infof("Evicting %s backend runner with model %s in %s mode",
 				r.backend, r.model, r.mode,
 			)
@@ -210,7 +217,10 @@ func (l *loader) Unload(ctx context.Context, unload UnloadRequest) int {
 			return l.evict(false)
 		} else {
 			for _, model := range unload.Models {
-				l.evictRunner(unload.Backend, model)
+				// Evict both, completion and embedding models. We should consider
+				// accepting a mode parameter in unload requests.
+				l.evictRunner(unload.Backend, model, inference.BackendModeCompletion)
+				l.evictRunner(unload.Backend, model, inference.BackendModeEmbedding)
 			}
 			return len(l.runners)
 		}
@@ -364,6 +374,8 @@ func (l *loader) load(ctx context.Context, backendName, model string, mode infer
 
 	// Loop until we can satisfy the request or an error occurs.
 	for {
+		slot := -1
+
 		// If loads are disabled, then there's nothing we can do.
 		if !l.loadsEnabled {
 			return nil, errLoadsDisabled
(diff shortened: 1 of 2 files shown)


pkg/inference/scheduling/loader.go
```diff
@@ -372,9 +372,15 @@ func (l *loader) load(ctx context.Context, backendName, model string, mode infer
 		// See if we can satisfy the request with an existing runner.
 		existing, ok := l.runners[runnerKey{backendName, model, mode}]
 		if ok {
-			l.references[existing] += 1
-			l.timestamps[existing] = time.Time{}
-			return l.slots[existing], nil
+			select {
+			case <-l.slots[existing].done:
```
doringeman, line 388: I think it'd make sense to also run `l.evictRunner(backendName, model`) so we don't have to evict all runners in order to find a free slot. WDYT?

p1-0tr: Yep, that makes sense.

pkg/inference/scheduling/loader.go
```diff
@@ -372,9 +372,16 @@ func (l *loader) load(ctx context.Context, backendName, model string, mode infer
...
+				l.log.Warnf("Will reload defunct %s runner for %s. Runner error: %s.", backendName, model,
+					l.slots[existing].err)
+				l.evictRunner(backendName, model)
```
doringeman:
```suggestion
				// Reset the reference count to zero so that we can evict the runner and then start a new one.
				l.references[existing] = 0
				l.evictRunner(backendName, model)
```

p1-0tr: Makes sense. Though I wonder if it would not be safer to let the reference counting work normally, issue and idle check here, and expand the idle check logic to look for defunct or stale runners. WDYT?

doringeman: > expand the idle check logic to look for defunct or stale runners

I like this!

Although, in this specific case, this code which comes right after the code you're changing will evict all (1, currently, but still) runners if all the slots are full and the current one that's attempted to be loaded is defunct and not clean up, right?
```
// If there's not sufficient memory or all slots are full, then try
// evicting unused runners.
if memory > l.availableMemory || len(l.runners) == len(l.slots) {
	l.evict(false)
}
```

p1-0tr: I'm pretty sure forcing the refcount to 0 does put us at a risk of `panic`ing in `loader.release`. I've opted not to force the refcount to 0, and added logic in `evict` to remove defunct runners.

pkg/inference/scheduling/loader.go
```diff
@@ -162,7 +162,13 @@ func (l *loader) evict(idleOnly bool) int {
...
 		unused := l.references[slot] == 0
 		idle := unused && now.Sub(l.timestamps[slot]) > runnerIdleTimeout
-		if unused && (!idleOnly || idle) {
+		defunct := false
+		select {
+		case <-l.slots[slot].done:
+			defunct = true
+		default:
+		}
+		if unused && (!idleOnly || idle || defunct) {
```
xenoscopic, line 172: This chunk looks good, I would just update the doc comment for `evict` to reflect that it also evicts defunct runners if possible.

pkg/inference/scheduling/loader.go
xenoscopic: I agree that we can't force the refcount to `0` here.

The bigger issue I see with the new logic is that `evictRunner` in this case might not actually evict if there's a non-zero reference count for the defunct runner (e.g. a client that hasn't realized its backend is defunct yet).  The problem is that this code would then continue and override the `l.runners` entry for `runnerKey{backend, model, mode}` with a newly created runner, so when that hypothetical outstanding defunct runner is finally released, it will decrement the reference count for the new runner in `release` (since it uses the same key to look up the slot).

I think what I would do is put a label (say `WaitForChange:`) just above the last block of code in this loop (grep for "Wait for something to change") and then in the `case <-l.slots[existing].done:` path, I would `goto WaitForChange`.  Then, in `release`, add a check for `<-runner.done` and immediately evict if `l.references[slot] == 0`.  Because realistically any client using a defunct runner will find out quite quickly once the socket connection closes, which means the runner will be `release`'d quickly, which will call `broadcast` and break the waiting `load` call out of its waiting loop.

xenoscopic (changes requested):
I like the idea, but I think we'll need a slightly different approach.

pkg/inference/scheduling/loader.go
p1-0tr: Done :)

pkg/inference/scheduling/loader.go
p1-0tr: Ok, it took me a while to convince myself that `run()` and `load()` will play nice with this approach. Should be all good, though :)

pkg/inference/scheduling/loader.go
```diff
@@ -364,6 +371,8 @@ func (l *loader) load(ctx context.Context, backendName, model string, mode infer
 
 	// Loop until we can satisfy the request or an error occurs.
 	for {
+		slot := -1
```
xenoscopic, line 377: Curious on the need for relocating this variable - does the `goto` require it?

pkg/inference/scheduling/loader.go
```diff
@@ -455,13 +470,19 @@ func (l *loader) release(runner *runner) {
...
-	// start time and signal the idle checker.
+	// If the runner's reference count is now zero, then check if it is still
+	// active, and record now as its idle start time and signal the idle
+	// checker.
 	if l.references[slot] == 0 {
-		l.timestamps[slot] = time.Now()
 		select {
-		case l.idleCheck <- struct{}{}:
+		case <-runner.done:
+			l.evictRunner(runner.backend.Name(), runner.model)
```
xenoscopic: If we want to get REALLY pedantic, we could also add an optional `mode` specification to `evictRunner` so that eviction of a defunct runner in one mode doesn't force early eviction of an idle runner in another mode.  Most (all?) models aren't used bi-modally though, so maybe it's a non-issue.

xenoscopic (approved):
Looks good, just two thoughts:

pkg/inference/scheduling/loader.go
p1-0tr: Yep, the compiler does not like jumping over variable declarations.

pkg/inference/scheduling/loader.go
p1-0tr: Makes the semantics clearer, I like it :) done.

doringeman (approved):
LGTM!
I wonder if it makes sense to also change `idleCheckDuration` to check immediately if a defunct runner exists.

pkg/inference/scheduling/loader.go
p1-0tr: Ended up having to add an error handler on the proxy, where we can wait for the runner process to exit. Otherwise we race, and often end up `release`ing the runner in the gap between it closing its socket and the `done` channel being closed.

xenoscopic (approved):
Looks great to me!

xenoscopic:
@doringeman I don't know if one defunct runner would indicate a high likelihood of other defunct runners.

doringeman:
@xenoscopic I meant to modify `idleCheckDuration` so that when it's called randomly, it also verifies whether there's a defunct runner, and if so, returns a maximum value to trigger an immediate check.

xenoscopic:
@doringeman Sorry, I misunderstood.  If I'm following correctly now, then I'm definitely in agreement.  Instead of "maximum value" though, just `0` (to check immediately), right?

doringeman:
Yeah, `0` to check immediately, sorry.

doringeman:
https://github.com/docker/model-runner/pull/77

PR #69, opened 2025-06-05
Return error in case of runner crash

p1-0tr:
In case the runner crashes it would be nice to return an error to the
user. So, add an error handler on the proxy and use it to try to figure
out if the runner crashed and format the response error accordingly.

Based on: https://github.com/docker/model-runner/pull/68

### DIFF
--- a/pkg/inference/scheduling/runner.go
+++ b/pkg/inference/scheduling/runner.go
@@ -161,7 +161,6 @@ func run(
 			}
 		} else {
 			w.WriteHeader(http.StatusBadGateway)
-
 		}
 	}
 


pkg/inference/scheduling/runner.go
```diff
@@ -134,6 +135,28 @@ func run(
 		proxyLog:  proxyLog,
 	}
 
+	proxy.ErrorHandler = func(w http.ResponseWriter, req *http.Request, err error) {
```
xenoscopic: Since this is going to be sent back to OpenAI clients, I would try to structure the error responses in a format they can parse.  Documentation is a little sparse, but Google says the error format looks like:

```
{
  "error": {
    "message": "Invalid 'messages[1].content': string too long. Expected a string with maximum length 1048576, but got a string with length 1540820 instead.",
    "type": "invalid_request_error",
    "param": "messages[1].content",
    "code": "string_above_max_length"
  }
}
```

The only docs I can find for the API are https://platform.openai.com/docs/api-reference/responses-streaming/error

doringeman (commented):
How can I easily test this to see the error?

pkg/inference/scheduling/runner.go
```diff
@@ -143,11 +144,24 @@ func run(
...
+				if err == nil {
+					w.Header().Set("Content-Type", "application/json; charset=utf-8")
+					w.Write(errJson)
+				}
 				return
 			case <-time.After(30 * time.Second):
 			}
 		} else {
 			w.WriteHeader(http.StatusBadGateway)
+
```
doringeman:
```suggestion
```

xenoscopic (approved):

PR #72, opened 2025-06-10
Get remote models

ilopezluna:
Adds ability to query models from remote registries via `?remote=true` query parameter.

- Added registry client to model manager for remote operations
- Split model retrieval into local/remote handlers
- Made `Tags` field optional in JSON responses

`GET /models/{name}?remote=true` - Fetch model from remote registry

@doringeman We initially discussed creating a common type for [local and remote models](https://github.com/docker/model-distribution/blob/main/types/model.go), but I ultimately decided against it because I don’t see a strong need.

We use different clients to retrieve local vs. remote models (the distribution client for local and the registry client for remote), and each returns its own type, which makes sense given their responsibilities.

For example, in the case of local models, it's important to return a [Model](https://github.com/docker/model-distribution/blob/b377026db94a863fceaec1a3088f924e80825888/types/model.go#L7C6-L7C11) in order to provide `GGUFPath()`

### DIFF
--- a/pkg/inference/models/manager.go
+++ b/pkg/inference/models/manager.go
@@ -37,6 +37,8 @@ type Manager struct {
 	router *http.ServeMux
 	// distributionClient is the client for model distribution.
 	distributionClient *distribution.Client
+	// registryClient is the client for model registry.
+	registryClient *registry.Client
 }
 
 type ClientConfig struct {
@@ -65,12 +67,19 @@ func NewManager(log logging.Logger, c ClientConfig, allowedOrigins []string) *Ma
 		// respond to requests, but may return errors if the client is required.
 	}
 
+	// Create the model registry client.
+	registryClient := registry.NewClient(
+		registry.WithTransport(c.Transport),
+		registry.WithUserAgent(c.UserAgent),
+	)
+
 	// Create the manager.
 	m := &Manager{
 		log:                log,
 		pullTokens:         make(chan struct{}, maximumConcurrentModelPulls),
 		router:             http.NewServeMux(),
 		distributionClient: distributionClient,
+		registryClient:     registryClient,
 	}
 
 	// Register routes.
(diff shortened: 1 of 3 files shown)


pkg/inference/models/manager.go
```diff
@@ -218,6 +239,55 @@ func (m *Manager) handleGetModel(w http.ResponseWriter, r *http.Request) {
...
+
+	return ToModel(model)
+}
+
+func getRemoteModel(m *Manager, name string) (*Model, error) {
+	if m.registryClient == nil {
+		return nil, errors.New("registry client unavailable")
+	}
+
+	model, err := m.registryClient.Model(context.Background(), name)
```
doringeman: We can pass the request's context `r.Context()` and use it instead of `context.Background()`.

pkg/inference/models/manager.go
```diff
@@ -218,6 +239,55 @@ func (m *Manager) handleGetModel(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	return ToModel(model)
+}
+
+func getRemoteModel(m *Manager, name string) (*Model, error) {
+	if m.registryClient == nil {
+		return nil, errors.New("registry client unavailable")
+	}
+
```
doringeman, line 260:
```suggestion
	m.log.Infoln("Getting remote model:", name)
```
To log it as it's done for local models in ⬇️ ?
```
func (c *Client) GetModel(reference string) (types.Model, error) {
	c.log.Infoln("Getting model by reference:", reference)
```

pkg/inference/models/manager.go
```diff
@@ -218,6 +239,55 @@ func (m *Manager) handleGetModel(w http.ResponseWriter, r *http.Request) {
...
+	config, err := model.Config()
+	if err != nil {
+		return nil, err
+	}
+
+	apiModel := &Model{
+		ID:      id,
+		Tags:    make([]string, 0),
```
doringeman:
```suggestion
		Tags:    []string{name},
```
Is this worth it?

ilopezluna: I think it’s more appropriate to use an empty array, or even nil, since we don’t really know the tags of the remote model. Also, the name could include the registry or even be a digest, so we shouldn’t use it as a tag.

ilopezluna: Actually I think `nil` is the right value. I changed to use `nil` instead empty array. Let me know wdyt @doringeman 🙏

doringeman: Looks good, thank you! Tested again, all good! 🙌

PR #73, opened 2025-06-10
Allow specifying a target llama-server version

p1-0tr:
I've found being able to lock the `llama-server` version to be useful at times. If we want to merge this it may make sense to come up with a cleaner solution. LMK what you think.

### DIFF
--- a/main.go
+++ b/main.go
@@ -48,6 +48,11 @@ func main() {
 		llamacpp.ShouldUpdateServerLock.Unlock()
 	}
 
+	desiredSeverVersion, ok := os.LookupEnv("LLAMACPP_SERVER_VERSION")
+	if ok {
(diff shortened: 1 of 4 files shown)


doringeman (approved):
Sure! This allows testing new versions before tagging them as `latest` without any code change!

main.go
```diff
@@ -48,6 +48,11 @@ func main() {
 		llamacpp.ShouldUpdateServerLock.Unlock()
 	}
 
+	desiredSeverVersion, ok := os.LookupEnv("SERVER_VERSION")
```
xenoscopic: Only minor nit is that maybe we should prefix the environment variable with an engine spec, e.g. `LLAMACPP_SERVER_VERSION`, for the future where we have multiple backends.

PR #74, opened 2025-06-10
Rebuild ModelsManager and Scheduler routers with new CORS origins

doringeman:
Allow rebuilding the routers with new allowed origins for CORS.

### DIFF
--- a/pkg/inference/models/manager.go
+++ b/pkg/inference/models/manager.go
@@ -10,6 +10,7 @@ import (
 	"path"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/docker/model-distribution/distribution"
 	"github.com/docker/model-distribution/registry"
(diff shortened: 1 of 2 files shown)


pkg/inference/models/manager.go
```diff
@@ -100,6 +100,18 @@ func NewManager(log logging.Logger, c ClientConfig, allowedOrigins []string) *Ma
 	return m
 }
 
+func (m *Manager) RebuildRoutes(allowedOrigins []string) {
```
p1-0tr, line 106: Feels like something that should happen behind a lock, or is the consumer of this API supposed to lock?

pkg/inference/models/manager.go
```diff
@@ -26,6 +27,8 @@ const (
 	maximumConcurrentModelPulls = 2
 )
 
+var lock sync.Mutex
```
p1-0tr: Why not make it a member of `Manager`?

pkg/inference/models/manager.go
doringeman: You're right, added a lock in https://github.com/docker/model-runner/commit/784d6dfee96bee79de71212db4eb6cbb52dc2121 ([compare the force push changes](https://github.com/docker/model-runner/compare/7eeb62e089e8120585a6ecab8e55a4aa1044182a..784d6dfee96bee79de71212db4eb6cbb52dc2121)).

PR #76, opened 2025-06-11
Support runner configuration

p1-0tr:
We need to allow users to configure the model runtime. Whether to
control inference settings, or low-level llama.cpp specific settings.

In the interest of unblocking users quickly, this patch adds a very simple
mechanism to configure the runtime settings. A `_configure` endpoint is
added per-engine, and acceps POST requests to set context-size and raw
runtime CLI flags. Those settings will be applied to any run of a given
model, until unload is called for that model or model-runner is
terminated.

This is a temporary solution and therefore subject to change once a
design for specifying runtime settings is finalised.


pkg/inference/scheduling/scheduler.go
```diff
@@ -112,6 +112,8 @@ func (s *Scheduler) routeHandlers(allowedOrigins []string) map[string]http.Handl
 	m["GET "+inference.InferencePrefix+"/ps"] = s.GetRunningBackends
 	m["GET "+inference.InferencePrefix+"/df"] = s.GetDiskUsage
 	m["POST "+inference.InferencePrefix+"/unload"] = s.Unload
+	m["POST "+inference.InferencePrefix+"/{backend}/configure"] = s.Configure
+	m["POST "+inference.InferencePrefix+"/configure"] = s.Configure
```
xenoscopic: If we think this is an interim solution for the next couple weeks, I'd maybe rename these to `_configure`.

pkg/inference/backends/llamacpp/llamacpp.go
```diff
@@ -138,6 +139,15 @@ func (l *llamaCpp) Run(ctx context.Context, socket, model string, mode inference
 	}
 
 	args := l.config.GetArgs(modelPath, socket, mode)
+
+	if config != nil {
+		args = append(args, "--ctx-size", strconv.Itoa(int(config.ContextSize)))
+		// FIXME(p1-0tr): this needs to be parsed, to respect quoted values etc.
```
xenoscopic: https://github.com/mattn/go-shellwords generally works well.

pkg/inference/scheduling/scheduler.go
```diff
@@ -347,6 +349,44 @@ func (s *Scheduler) Unload(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	var runnerConfig inference.BackendConfiguration
+	runnerConfig.ContextSize = configureRequest.ContextSize
+	runnerConfig.RawFlags = configureRequest.RawRuntimeFlags
+
+	s.loader.setRunnerConfig(r.Context(), backend.Name(), configureRequest.Model, inference.BackendModeCompletion, runnerConfig)
+
+	w.WriteHeader(http.StatusOK)
```
xenoscopic:
```suggestion
	w.WriteHeader(http.StatusAccepted)
```
Might be more appropriate since 200 implies status in response body. 202 indicates async (i.e. won't know if it worked until backend starts).

xenoscopic (approved):
LGTM overall, hard to fully assess since we think this is sort of temporary.  I think my only comment of consequence would be to avoid baking in any APIs or code refactors that would be hard to undo if they don't serve the later design.

p1-0tr:
> I'm only a bit concerned about the UX when running multiple clients against the same model runner. It seems to me that changing the configuration of an already loaded model will silently ignore the change until the model is restarted. Is there somehting we could do to make this more apparent?

That's a fair point. We are shooting for a minimal solution, until we can get a proper design for handling configs dialed. Keeping that in mind, we could, for example return a `403` in case someone tries to configure an already running model.

fiam:
> > I'm only a bit concerned about the UX when running multiple clients against the same model runner. It seems to me that changing the configuration of an already loaded model will silently ignore the change until the model is restarted. Is there somehting we could do to make this more apparent?
> 
> That's a fair point. We are shooting for a minimal solution, until we can get a proper design for handling configs dialed. Keeping that in mind, we could, for example return a `403` in case someone tries to configure an already running model.

Should we include that change in this PR? It seems to me it would follow the principle of least surprise by failing loudly.

p1-0tr:
> Should we include that change in this PR? It seems to me it would follow the principle of least surprise by failing loudly.

Yes :) I pushed it a couple of minutes ago :)

pkg/inference/scheduling/scheduler.go
```diff
@@ -347,6 +350,60 @@ func (s *Scheduler) Unload(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	var runnerConfig inference.BackendConfiguration
+	runnerConfig.ContextSize = configureRequest.ContextSize
+	runnerConfig.RawFlags = rawFlags
+
+	if err := s.loader.setRunnerConfig(r.Context(), backend.Name(), configureRequest.Model, inference.BackendModeCompletion, runnerConfig); err != nil {
+		if err == errRunnerAlreadyActive {
+			w.WriteHeader(http.StatusForbidden)
```
fiam: tiny nit: log error in case we have to debug?

fiam:
> > Should we include that change in this PR? It seems to me it would follow the principle of least surprise by failing loudly.
> 
> Yes :) I pushed it a couple of minutes ago :)

Excellent, thank you!

pkg/inference/scheduling/scheduler.go
```diff
@@ -347,6 +350,61 @@ func (s *Scheduler) Unload(w http.ResponseWriter, r *http.Request) {
...
+	runnerConfig.ContextSize = configureRequest.ContextSize
+	runnerConfig.RawFlags = rawFlags
+
+	if err := s.loader.setRunnerConfig(r.Context(), backend.Name(), configureRequest.Model, inference.BackendModeCompletion, runnerConfig); err != nil {
+		s.log.Warnf("Failed to configure %s runner for %s: %s", backend.Name(), configureRequest.Model, err)
+		if err == errRunnerAlreadyActive {
+			w.WriteHeader(http.StatusForbidden)
+		} else {
+			w.WriteHeader(http.StatusInternalServerError)
+		}
```
doringeman: WDYT of writing the error in the body as well?
```suggestion
		if errors.Is(err, errRunnerAlreadyActive) {
			http.Error(w, err.Error(), http.StatusForbidden)
		} else {
			http.Error(w, err.Error(), http.StatusInternalServerError)
		}
```

pkg/inference/scheduling/scheduler.go
```diff
@@ -347,6 +350,61 @@ func (s *Scheduler) Unload(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	var runnerConfig inference.BackendConfiguration
+	runnerConfig.ContextSize = configureRequest.ContextSize
+	runnerConfig.RawFlags = rawFlags
+
+	if err := s.loader.setRunnerConfig(r.Context(), backend.Name(), configureRequest.Model, inference.BackendModeCompletion, runnerConfig); err != nil {
+		s.log.Warnf("Failed to configure %s runner for %s: %s", backend.Name(), configureRequest.Model, err)
+		if errors.Is(err, errRunnerAlreadyActive) {
+			http.Error(w, err.Error(), http.StatusForbidden)
```
xenoscopic:
```suggestion
			http.Error(w, err.Error(), http.StatusConflict)
```
Putting on my "fun at parties" hat, 409 might _technically_ be more appropriate: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/409

pkg/inference/scheduling/scheduler.go
```diff
@@ -347,6 +350,61 @@ func (s *Scheduler) Unload(w http.ResponseWriter, r *http.Request) {
...
+	runnerConfig.ContextSize = configureRequest.ContextSize
+	runnerConfig.RawFlags = rawFlags
+
+	if err := s.loader.setRunnerConfig(r.Context(), backend.Name(), configureRequest.Model, inference.BackendModeCompletion, runnerConfig); err != nil {
+		s.log.Warnf("Failed to configure %s runner for %s: %s", backend.Name(), configureRequest.Model, err)
+		if err == errRunnerAlreadyActive {
+			w.WriteHeader(http.StatusConflict)
+		} else {
+			w.WriteHeader(http.StatusInternalServerError)
+		}
```
doringeman:
```suggestion
		if errors.Is(err, errRunnerAlreadyActive) {
			http.Error(w, err.Error(), http.StatusConflict)
		} else {
			http.Error(w, err.Error(), http.StatusInternalServerError)
		}
```
It looks like you forgot to pull before force pushing.

p1-0tr: I knew I should not have pushed `commit suggestion` yesterday 🤦

p1-0tr: This time I updated my local copy immediately.

PR #77, opened 2025-06-12
loader: trigger idle runner eviction immediately if a runner is defunct

doringeman:
Modify `idleCheckDuration` so that when it's called randomly, it also verifies whether there's a defunct runner, and if so, returns `0` to trigger an immediate check.


xenoscopic (approved):
Great idea, sorry for my initial confusion.  My brain is running on the 1B parameter model at the moment 🫠

PR #81, opened 2025-06-13
bump model-distribution

ilopezluna:


xenoscopic (approved):

pkg/inference/models/manager_test.go
```diff
@@ -119,7 +153,7 @@ func TestPullModel(t *testing.T) {
 				r.Header.Set("Accept", tt.acceptHeader)
 			}
 
-			w := httptest.NewRecorder()
+			w := newMutexResponseRecorder()
```
ekcasey, line 156: I don't think we can assume the real `http.ResponseWriter` is thread safe and we might be papering over real issues with this change to use a thread-safe version in the tests. It looks to me like the concurrent writes are happening within the pull implementation.

My best guess is this is that this is appearing now b/c with the [per-layer progress reporting](https://github.com/docker/model-distribution/pull/80) we are no longer [waiting](https://github.com/docker/model-distribution/blob/a9b8592eff182754d6a54f909da4416739097411/internal/progress/reporter.go#L86-L89) for the progress writing from one layer to complete before we start writing progress for the next layer.

ilopezluna:
Closing because the race is in the pull implementation: https://github.com/docker/model-distribution/pull/89

PR #85, opened 2025-06-16
Create home folder

ilopezluna:


p1-0tr (approved):
LGTM

doringeman:
Should we change the default MODELS_PATH as well?

xenoscopic:
I'd opt to keep the current `MODELS_PATH` just because it makes for a simpler syntax for end-users using `-v`.

xenoscopic (approved):

PR #88, opened 2025-06-18
Bump model-distribution

ilopezluna:


xenoscopic (approved):
LGTM, only nit is maybe run a `go mod tidy` to prune the old `go.sum` entries.

PR #93, opened 2025-06-25
Respect context size from model config

ekcasey:
* Respects context-size defined in the model artifact (see https://github.com/docker/model-distribution/pull/92). This takes precedence over backend config.
* Small refactoring of arg generation and testing

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.23.7
 require (
 	github.com/containerd/containerd/v2 v2.0.4
 	github.com/containerd/platforms v1.0.0-rc.1
-	github.com/docker/model-distribution v0.0.0-20250627152504-c0c68acaabd5
+	github.com/docker/model-distribution v0.0.0-20250627163720-aff34abcf3e0
 	github.com/google/go-containerregistry v0.20.3
 	github.com/jaypipes/ghw v0.16.0
 	github.com/mattn/go-shellwords v1.0.12
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.23.7
 require (
 	github.com/containerd/containerd/v2 v2.0.4
 	github.com/containerd/platforms v1.0.0-rc.1
-	github.com/docker/model-distribution v0.0.0-20250618082521-fb5c8332c857
+	github.com/docker/model-distribution v0.0.0-20250627152504-c0c68acaabd5
 	github.com/google/go-containerregistry v0.20.3
 	github.com/jaypipes/ghw v0.16.0
 	github.com/mattn/go-shellwords v1.0.12
--- a/go.sum
+++ b/go.sum
@@ -38,10 +38,8 @@ github.com/docker/distribution v2.8.3+incompatible h1:AtKxIZ36LoNK51+Z6RpzLpddBi
 github.com/docker/distribution v2.8.3+incompatible/go.mod h1:J2gT2udsDAN96Uj4KfcMRqY0/ypR+oyYUYmja8H+y+w=
 github.com/docker/docker-credential-helpers v0.8.2 h1:bX3YxiGzFP5sOXWc3bTPEXdEaZSeVMrFgOr3T+zrFAo=
 github.com/docker/docker-credential-helpers v0.8.2/go.mod h1:P3ci7E3lwkZg6XiHdRKft1KckHiO9a2rNtyFbZ/ry9M=
-github.com/docker/model-distribution v0.0.0-20250618082521-fb5c8332c857 h1:2IvvpdPZvpNn06+RUh5DC5O64dnrKjdsBKCMrzR5QTk=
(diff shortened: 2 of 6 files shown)


go.mod
```diff
@@ -5,7 +5,7 @@ go 1.23.7
 require (
 	github.com/containerd/containerd/v2 v2.0.4
 	github.com/containerd/platforms v1.0.0-rc.1
-	github.com/docker/model-distribution v0.0.0-20250618082521-fb5c8332c857
```
ekcasey, line 8: **** DO NOT MERGE until this ref points to a commit on the main branch ****

go.sum
```diff
@@ -40,6 +40,8 @@ github.com/docker/docker-credential-helpers v0.8.2 h1:bX3YxiGzFP5sOXWc3bTPEXdEaZ
 github.com/docker/docker-credential-helpers v0.8.2/go.mod h1:P3ci7E3lwkZg6XiHdRKft1KckHiO9a2rNtyFbZ/ry9M=
 github.com/docker/model-distribution v0.0.0-20250618082521-fb5c8332c857 h1:2IvvpdPZvpNn06+RUh5DC5O64dnrKjdsBKCMrzR5QTk=
 github.com/docker/model-distribution v0.0.0-20250618082521-fb5c8332c857/go.mod h1:dThpO9JoG5Px3i+rTluAeZcqLGw8C0qepuEL4gL2o/c=
+github.com/docker/model-distribution v0.0.0-20250627152504-c0c68acaabd5 h1:4qs7OwrJJQE1de5kbHg83gm5rmCRQzAwAp8gG/3cLY8=
+github.com/docker/model-distribution v0.0.0-20250627152504-c0c68acaabd5/go.mod h1:dThpO9JoG5Px3i+rTluAeZcqLGw8C0qepuEL4gL2o/c=
```
xenoscopic: nit: `go mod tidy` will prune the older `github.com/docker/model-distribution` entries.

pkg/inference/backends/llamacpp/llamacpp_config.go
```diff
@@ -33,10 +35,20 @@ func NewDefaultLlamaCppConfig() *Config {
...
 
 // GetArgs implements BackendConfig.GetArgs.
-func (c *Config) GetArgs(modelPath, socket string, mode inference.BackendMode) []string {
+func (c *Config) GetArgs(model types.Model, socket string, mode inference.BackendMode, config *inference.BackendConfiguration) ([]string, error) {
 	// Start with the arguments from LlamaCppConfig
 	args := append([]string{}, c.Args...)
 
+	modelPath, err := model.GGUFPath()
+	if err != nil {
+		return nil, fmt.Errorf("get gguf path: %v", err)
```
xenoscopic:
```suggestion
		return nil, fmt.Errorf("get gguf path: %w", err)
```

pkg/inference/backends/llamacpp/llamacpp_config.go
```diff
@@ -33,10 +35,20 @@ func NewDefaultLlamaCppConfig() *Config {
...
+	}
+
+	modelCfg, err := model.Config()
+	if err != nil {
+		return nil, fmt.Errorf("get model config: %v", err)
```
xenoscopic:
```suggestion
		return nil, fmt.Errorf("get model config: %w", err)
```

pkg/inference/backends/llamacpp/llamacpp_config.go
```diff
@@ -45,7 +57,20 @@ func (c *Config) GetArgs(modelPath, socket string, mode inference.BackendMode) [
...
+		args = append(args, "--ctx-size", strconv.FormatUint(*modelCfg.ContextSize, 10))
+	}
+
+	// Add arguments from backend config
+	if config != nil {
+		if config.ContextSize > 0 && !containsArg(args, "--ctx-size") {
+			args = append(args, "--ctx-size", fmt.Sprintf("%d", config.ContextSize))
+		}
+		args = append(args, config.RuntimeFlags...)
+	}
```
xenoscopic: I would unify the number-to-string conversion; `strconv.Itoa` is probably the most efficient (vs. `strconv.FormatUint` or `fmt.Sprintf`).

ekcasey: I can use `strconv.FormatInt` and `strconv.FormatUint` for the backend config and artifact config respectively. `strconv.Itoa` accepts an `int` type rather than an `int64`. I was assuming we chose int64 in the backend config for a reason and wouldn't want to risk losing precision.

go.sum
ekcasey: Fixed with the update to point at `main` branch

pkg/inference/backends/llamacpp/llamacpp_config.go
xenoscopic: Ah, roger that.

xenoscopic (approved):

PR #94, opened 2025-06-26
scheduler: move OpenAI recording after runner load/release

doringeman:
If the runner is defunct, then `loader.load` will evict it, which means it'll also remove the recorded OpenAI requests/responses for that model. Hence, record the request after `loader.load` returns so it doesn't get removed in the loading process.

Without this patch, following these steps will result in a lost pair of request-response, the first one after the runner exited/was killed and had to be evicted because it was defunct.
```
$ MODEL_RUNNER_PORT=8080 make run # in a separate terminal

$ MODEL_RUNNER_HOST=http://localhost:8080 docker model run ai/smollm2 hi

$ curl -s http://localhost:8080/engines/requests\?model\=ai/smollm2 # all good, 1 record

$ pkill com.docker.llama-server

$ MODEL_RUNNER_HOST=http://localhost:8080 docker model run ai/smollm2 hi

$ # See `ERRO[0011] Model ai/smollm2 not found in records - 200` in the terminal where model-runner is running.
# This is because the request was recorded, then `loader.load` removed it, 
and then via a deferred function we attempted to record the response, but no matching request was found.
# Hence, we lost the first pair of request-response.

$ curl -s http://localhost:8080/engines/requests\?model\=ai/smollm2
No records found for model 'ai/smollm2'
```

### DIFF
--- a/pkg/inference/scheduling/scheduler.go
+++ b/pkg/inference/scheduling/scheduler.go
@@ -238,14 +238,6 @@ func (s *Scheduler) handleOpenAIInference(w http.ResponseWriter, r *http.Request
 		s.tracker.TrackModel(model)
 	}
 
-	// Record the request in the OpenAI recorder.
-	recordID := s.openAIRecorder.RecordRequest(request.Model, r, body)
-	w = s.openAIRecorder.NewResponseRecorder(w)
-	defer func() {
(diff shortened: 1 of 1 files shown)


p1-0tr (approved):
LGTM

xenoscopic (approved):

PR #95, opened 2025-06-26
Allow configuration through argument list (in addition to string)

xenoscopic:
We'll allow runtime flag configuration via either an explicit, pre-parsed list (via `runtime-flags`) or via a string (via `raw-runtime-flags`).  This will support the Compose `models:` spec as well as the Compose `model` provider, respectively.

### DIFF
--- a/pkg/inference/backend.go
+++ b/pkg/inference/backend.go
@@ -30,8 +30,8 @@ func (m BackendMode) String() string {
 }
 
 type BackendConfiguration struct {
-	ContextSize int64    `json:"context_size,omitempty"`
-	RawFlags    []string `json:"flags,omitempty"`
+	ContextSize  int64    `json:"context-size,omitempty"`
+	RuntimeFlags []string `json:"runtime-flags,omitempty"`
 }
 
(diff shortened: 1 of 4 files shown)


pkg/inference/backend.go
```diff
@@ -30,8 +30,8 @@ func (m BackendMode) String() string {
 }
 
 type BackendConfiguration struct {
-	ContextSize int64    `json:"context_size,omitempty"`
-	RawFlags    []string `json:"flags,omitempty"`
+	ContextSize  int64    `json:"context-size,omitempty"`
+	RuntimeFlags []string `json:"runtime-flags,omitempty"`
```
xenoscopic, line 34: I don't know if this JSON nomenclature is used anywhere (I couldn't find it), but I've tried to make it consistent before it becomes used anywhere.

xenoscopic:
I don't think there's a strong Docker convention on the JSON keys (Docker API, Compose, et al. all look different).  Right now we've already released provider support with config in Docker CE, so I think we have to leave the keys as is for now 😞

PR #98, opened 2025-07-02
Use modelID instead of model tag

ilopezluna:
For actions with models:
- chat
- record req/res
- load
- evict

Use the model ID (digest) instead of the tag.
This has several benefits, like we don't need to unload/load the same model if user specifies different tag.

Also this PR include the actual reference used when running a model, to use it when doing `docker model ps` to mimic the behavior of `docker ps`, if you run an image with multiple tags, the `docker ps` command will show you the tag used

### DIFF
--- a/pkg/inference/scheduling/loader.go
+++ b/pkg/inference/scheduling/loader.go
@@ -214,8 +214,8 @@ func (l *loader) evict(idleOnly bool) int {
 		default:
 		}
 		if unused && (!idleOnly || idle || defunct) {
-			l.log.Infof("Evicting %s backend runner with model %s in %s mode",
-				r.backend, r.modelID, r.mode,
+			l.log.Infof("Evicting %s backend runner with model %s (%s) in %s mode",
+				r.backend, r.modelID, runnerInfo.modelRef, r.mode,
 			)
 			l.slots[runnerInfo.slot].terminate()
 			l.slots[runnerInfo.slot] = nil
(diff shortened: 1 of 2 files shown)


pkg/metrics/openai_recorder.go
```diff
@@ -71,46 +74,50 @@ func (r *OpenAIRecorder) SetConfigForModel(model string, config *inference.Backe
...
 			Config:  inference.BackendConfiguration{},
 		}
 	}
 
-	r.records[model].Config = *config
+	r.records[modelID].Config = *config
 }
 
 func (r *OpenAIRecorder) RecordRequest(model string, req *http.Request, body []byte) string {
+	modelID := r.modelManager.ResolveModelID(model)
```
xenoscopic, line 93: Does using the model ID in the `OpenAIRecorder` mean that the GUI that displays these requests will also display the SHA256?  Or do we resolve it back to the "friendly" name?  (Or something like `ai/gemma3@sha256:a484b...`?)

pkg/inference/scheduling/loader.go
```diff
@@ -224,17 +232,17 @@ func (l *loader) evict(idleOnly bool) int {
 // It returns the number of remaining runners.
 func (l *loader) evictRunner(backend, model string, mode inference.BackendMode) int {
 	allBackends := backend == ""
-	for r, slot := range l.runners {
-		unused := l.references[slot] == 0
-		if unused && (allBackends || r.backend == backend) && r.model == model && r.mode == mode {
+	for r, runnerInfo := range l.runners {
+		unused := l.references[runnerInfo.slot] == 0
+		if unused && (allBackends || r.backend == backend) && r.modelID == model && r.mode == mode {
 			l.log.Infof("Evicting %s backend runner with model %s in %s mode",
```
xenoscopic: It would be useful if the logging statements in `loader.go` and `scheduler.go` retained the model ref as well (maybe in a standardized format like `ai/<modelRef>@sha256:<modelID>`).  It would make debugging significantly easier.

pkg/metrics/openai_recorder.go
ilopezluna: Currently the UI does not uses/show tags when using a model or inspecting a model. Behind the scenes uses the model ID but the UI does not show that.
With @doringeman we mentioned that we are going to need a `ResolveModelTags(model reference)` to also show all the available tags eventually of a local model.

pkg/inference/scheduling/loader.go
ilopezluna: @xenoscopic update [here](https://github.com/docker/model-runner/pull/98/commits/a23dc5e5142d03170b665748e13d2f41eaf82f86)

doringeman (approved):

PR #100, opened 2025-07-04
models: return untagged model information on delete

doringeman:
Return untagged model information on `DELETE <inference-prefix>/models/{name}` requests.

Bump model-distribution to https://github.com/docker/model-distribution/commit/08f8ace513ff803b147a968d1ff36d39a1e536ef for the change of `m.distributionClient.DeleteModel`.

### DIFF
--- a/pkg/inference/models/manager.go
+++ b/pkg/inference/models/manager.go
@@ -333,7 +333,8 @@ func (m *Manager) handleDeleteModel(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	if err := m.distributionClient.DeleteModel(r.PathValue("name"), force); err != nil {
+	out, err := m.distributionClient.DeleteModel(r.PathValue("name"), force)
+	if err != nil {
 		if errors.Is(err, distribution.ErrModelNotFound) {
 			http.Error(w, err.Error(), http.StatusNotFound)
 			return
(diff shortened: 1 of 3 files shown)


pkg/inference/models/manager.go
```diff
@@ -346,6 +347,11 @@ func (m *Manager) handleDeleteModel(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
+
+	w.Header().Set("Content-Type", "text/plain")
+	if _, err := w.Write([]byte(out)); err != nil {
+		http.Error(w, fmt.Sprintf("error writing response: %v", err), http.StatusInternalServerError)
+	}
```
xenoscopic: I missed the review window on docker/model-distribution#104, but would it maybe make more sense to have `DeleteModel` return some sort of structure (e.g. `[]string`) and then handle the formatting at the API and CLI layers?  E.g. the corresponding `DELETE /images` response uses a JSON structure to format the results: https://docs.docker.com/reference/api/engine/version/v1.51/#tag/Image/operation/ImageDelete

doringeman: Done in https://github.com/docker/model-distribution/pull/107, thanks!

pkg/inference/models/manager.go
```diff
@@ -346,6 +347,11 @@ func (m *Manager) handleDeleteModel(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
+
+	w.Header().Set("Content-Type", "application/json")
+	if err := json.NewEncoder(w).Encode(resp); err != nil {
+		http.Error(w, fmt.Sprintf("error writing response: %v", err), http.StatusInternalServerError)
```
xenoscopic, line 353: I'm not sure this'll do anything at this point if the encode fails mid-encode, but can't hurt I suppose.

xenoscopic (approved):

pkg/inference/models/manager.go
doringeman: I guess it at least overwrites the previously set `"Content-Type", "application/json"` and send back an error.

PR #101, opened 2025-07-08
metrics: include current request's User-Agent

doringeman:
Include current request's User-Agent, appended to the `Tracker`'s.
Also track model on `/_configure`.

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -60,32 +60,38 @@ func NewTracker(httpClient *http.Client, log logging.Logger, userAgent string, d
 	}
 }
 
-func (t *Tracker) TrackModel(model types.Model) {
+func (t *Tracker) TrackModel(model types.Model, userAgent string) {
 	if t.doNotTrack {
 		return
 	}
 
-	go t.trackModel(model)
+	go t.trackModel(model, userAgent)
 }
 
-func (t *Tracker) trackModel(model types.Model) {
+func (t *Tracker) trackModel(model types.Model, userAgent string) {
 	tags := model.Tags()
 	t.log.Debugln("Tracking model:", tags)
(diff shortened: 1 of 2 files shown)


pkg/metrics/metrics.go
```diff
@@ -60,32 +61,35 @@ func NewTracker(httpClient *http.Client, log logging.Logger, userAgent string, d
...
-	); err != nil {
-		t.log.Debugf("Manifest does not exist or error occurred: %v\n", err)
-		return
+	for _, tag := range tags {
+		ref, err := name.ParseReference(tag)
+		if err != nil {
+			t.log.Errorf("Error parsing reference: %v\n", err)
+			return
+		}
+		ua := strings.Join([]string{t.userAgent, userAgent}, " ")
```
xenoscopic:
```suggestion
		ua := t.userAgent
		if userAgent != "" {
			ua += " " + userAgent
		}
```
nit: avoid trailing space if user agent is unspecified
note: haven't tested this suggestion

xenoscopic (approved):

pkg/metrics/metrics.go
```diff
@@ -60,32 +60,38 @@ func NewTracker(httpClient *http.Client, log logging.Logger, userAgent string, d
...
+	for _, tag := range tags {
+		ref, err := name.ParseReference(tag)
+		if err != nil {
+			t.log.Errorf("Error parsing reference: %v\n", err)
+			return
+		}
+		ua := t.userAgent
+		if userAgent != "" {
+			ua += " " + userAgent
+		}
```
xenoscopic: Actually one more thought - this could be done outside the loop.

doringeman: Ha, of course! 😅

PR #106, opened 2025-07-11
Implement basic memory estimation in scheduler

p1-0tr:
First pass implementation of memory estimation logic in model scheduler. This change heavily relies on gguf-parser-go to calculate estimated peak memory requirement for running inference with a given model. It adds GetRequiredMemoryForModel() to the Backend interface to allow each backend to deal with its config and calculate required memory usage based on it.

Before merging:

- [x] Add support for windows (CUDA)
- [x] Add support for Linux (CUDA)
- [x] Estimate RAM usage, especially when GPU offload is not configured
- [x] Extend loader to keep track of RAM and VRAM (for now one VRAM space)

### DIFF
--- a/pkg/inference/scheduling/loader.go
+++ b/pkg/inference/scheduling/loader.go
@@ -141,7 +141,7 @@ func newLoader(
 		vramSize = 1
 		log.Warnf("Could not read VRAM size: %s", err)
 	} else {
-		log.Infof("Running on system with %dMB VRAM", vramSize/1022/1024)
+		log.Infof("Running on system with %dMB VRAM", vramSize/1024/1024)
 	}
 	ramSize := uint64(1)
 	hostInfo, err := sysinfo.Host()


pkg/inference/scheduling/metal.m
```diff
@@ -0,0 +1,13 @@
+// +build darwin
+
+#include <Metal/Metal.h>
+
+#include "metal.h"
+
+size_t getVRAMSize() {
+    id<MTLDevice> device = MTLCreateSystemDefaultDevice();
+    if (device) {
+        return [device recommendedMaxWorkingSetSize];
```
xenoscopic:
```suggestion
        size_t result = [device recommendedMaxWorkingSetSize];
        [device release];
        return result;
```
It may not matter here since `MTLCreateSystemDefaultDevice` probably returns a singleton, but with the [create rule](https://stackoverflow.com/a/60782204), I **_think_** you're responsible for `[device release]`.  Not 100% sure I've nailed the syntax in my suggestion.

pkg/inference/scheduling/loader.go
```diff
@@ -390,14 +380,16 @@ func (l *loader) load(ctx context.Context, backendName, model string, mode infer
...
-	memory := uint64(1)
+	var runnerConfig *inference.BackendConfiguration
+	if rc, ok := l.runnerConfigs[runnerKey{backendName, model, mode}]; ok {
+		runnerConfig = &rc
+	}
+	reqMemory, err := backend.GetRequiredMemoryForModel(model, runnerConfig)
+	if err != nil {
+		return nil, err
+	}
+	memory := reqMemory.VRAM // FIXME(p1-0tr): extend loader to reason about RAM and VRAM (and multiple VRAM sets in the future)
```
xenoscopic: We could probably adjust the logic to do dual comparisons already, i.e. convert `totalMemory` to `totalRAM` and `totalVRAM` and then change the conditions below to something like `if reqMemory.RAM > l.totalRAM || reqMemory.VRAM > l.totalVRAM`.  If we just leave `RAM` and `totalRAM` set to `0` for now, that's fine, but it would be nice to have the logic in place once we do start querying those.

pkg/inference/backends/llamacpp/llamacpp.go
```diff
@@ -213,3 +215,38 @@ func (l *llamaCpp) GetDiskUsage() (int64, error) {
...
+		return nil, err
+	}
+	mdlGguf, err := parser.ParseGGUFFile(mdlPath)
+	if err != nil {
+		return nil, err
+	}
+	mdlConfig, err := mdl.Config()
+	if err != nil {
+		return nil, err
+	}
```
xenoscopic: A bit more error wrapping might help here since there are so many failure modes.

Dockerfile
```diff
@@ -27,7 +27,7 @@ COPY --link . .
 # Build the Go binary (static build)
 RUN --mount=type=cache,target=/go/pkg/mod \
     --mount=type=cache,target=/root/.cache/go-build \
-    CGO_ENABLED=0 GOOS=linux go build -ldflags="-s -w" -o model-runner ./main.go
+    CGO_ENABLED=1 GOOS=linux go build -ldflags="-s -w" -o model-runner ./main.go
```
xenoscopic, line 30: For Windows and Linux, could we move the memory querying into the `nv-gpu-info` command that we ship alongside the `llama.cpp` backend?  Both of those platforms have terrible error reporting when dynamic linking fails to resolve a library, so I worry it might be confusing to users who accidentally enable GPU support but don't have the NVIDIA libraries available.

pkg/inference/scheduling/memory_windows.go
```diff
@@ -0,0 +1,16 @@
+package scheduling
+
+/*
+#include "nvapi.h"
+*/
+import "C"
+import "errors"
+
+// getVRAMSize returns total system GPU memory in bytes
+func getVRAMSize() (uint64, error) {
```
xenoscopic: For Windows, we also need to consider the OpenCL backend.  I don't think OpenCL has an analogous API for querying VRAM, but since we only support Adreno GPUs with shared memory, maybe we can just use the whole system memory (or some fraction thereof) as a proxy for the available GPU memory?

xenoscopic (commented):
Looking good so far!

pkg/inference/scheduling/loader.go
xenoscopic: OTOH, if we have to do really complex memory topologies later, maybe the comparison will have to be factored out behind some `MemoryComparer` interface.  So if you want to leave it as just a single comparison for now, that's probably fine.

p1-0tr: I think adding logic for keeping track of RAM usage for CPU-only cases, would make most sense in the first pass implementation. And later we can refine it to handle more complex topologies, and track utilisation in more detail.

xenoscopic: Makes sense to me.  Forgot that we need to handle the CPU case too.  So I'll let you be the judge.

Dockerfile
p1-0tr: I'm currently modifying the code to do that on Windows. For linux I was going to stick to the `dlopen` route, as I'm operating under the assumption that we are running in a container most of the time anyway (so the environment and library resolution should be pretty predictable most of the time).

pkg/inference/scheduling/memory_windows.go
p1-0tr: Yep, win arm64 is a bit problematic. I was considering querying available VRAM through DX or Vulkan (there shouldn't be a difference between VRAM available to OpenCL or regular graphics stacks) but that may be overkill. Alternatively we could 🙈 and stick to a single model limit (which is a decent estimate anyway, maybe apart from `smollm2`) on Adreno for now.

xenoscopic: I'm okay with sticking to the single model limit there for now.  That case is more complicated as well because the GPU only supports certain quantizations and most models will land on the CPU.  But if we can keep the logic within the same formalism (e.g. have `getVRAMSize()` return `1` there and `GetRequiredMemoryForModel()` return `1`), then we can work on it iteratively later.

pkg/inference/backends/llamacpp/llamacpp.go
```diff
@@ -213,3 +215,50 @@ func (l *llamaCpp) GetDiskUsage() (int64, error) {
...
+
+	contextSize := GetContextSize(&mdlConfig, config)
+
+	// FIXME(p1-0tr): for now assume we are running on GPU (single one) - Devices[1];
+	// sum up weights + kv cache + context for an estimate of total GPU memory needed
+	// while running inference with the given model
+	estimate := mdlGguf.EstimateLLaMACppRun(parser.WithLLaMACppContextSize(int32(contextSize)),
+		// FIXME(p1-0tr): add logic for resolving other param values, instead of hardcoding them
+		parser.WithLLaMACppLogicalBatchSize(2048),
+		parser.WithLLaMACppOffloadLayers(100)) // FIXME(p1-0tr): need to set this to 0 for CPU only case
```
p1-0tr: This `FIXME` still needs to be resolved (it's slightly annoying because we don't have a straightforward way of telling if we are running with or without GPU acceleration at this point)

xenoscopic: Can we just assume it on `darwin` and replicate the logic in `ensureLatestLlamaCpp` on Windows otherwise to determine if we think we're on the GPU?

p1-0tr: The most problematic case is Linux, since there we bake the server into the `model-runner` image. But I think `llama-server --list-devices` may be a good way to work around this (it should return an empty list if it is a CPU-only build)

xenoscopic: That or check for the NVIDIA environment variables that the CTK sets, e.g. `NVIDIA_VISIBLE_DEVICES`

p1-0tr: That's my fallback plan (not first choice, because it does not cover the case of someone running a CPU-only `llama-server` build in a container with GPUs passed through to it)

pkg/gpuinfo/nvidia.c
```diff
@@ -0,0 +1,71 @@
+// +build linux
```
xenoscopic: minor nit: I'd recommend using the newer `//go:build` directives here and elsewhere.

pkg/inference/backends/llamacpp/llamacpp.go
```diff
@@ -213,3 +222,88 @@ func (l *llamaCpp) GetDiskUsage() (int64, error) {
...
+	var vram uint64
+	if len(estimate.Devices) > 1 {
+		vram = uint64(estimate.Devices[1].Weight.Sum() + estimate.Devices[1].KVCache.Sum() + estimate.Devices[1].Computation.Sum())
+	}
+
+	if runtime.GOOS == "windows" && runtime.GOARCH == "arm64" {
+		if mdlConfig.Quantization == "Q4_0" {
+			// TODO(p1-0tr): For now on windows/arm64 stick to the old behaviour, of allowing
+			// one model at a time. This WA requires gpuinfo.GetVRAMSize to return 1.
+			vram = 1
```
xenoscopic: If there's potentially a risk of other quantizations besides `Q4_0` becoming supported on the GPU without our intervention, maybe we should set vram requirements to `1` ubiquitously on `windows`/`arm64` (in the same way that `GetVRAMSize` returns `1` ubiquitously) regardless of the quantization?  Admittedly I need to think this logic through a bit and review a bit more.

pkg/inference/scheduling/loader.go
```diff
@@ -132,20 +145,27 @@ func newLoader(
...
-	//
-	// HACK: On GPU-enabled cloud engines, we'll bump this to 2. We can remove
-	// this once we have VRAM estimation.
-	totalMemory := uint64(1)
-	if isGPUEnabledCloudEnvironment {
-		totalMemory = 2
+	// TODO(p1-0tr): improve error handling
+	vramSize, err := gpuInfo.GetVRAMSize()
+	if err != nil {
+		log.Warnf("Could not read VRAM size: %s", err)
```
xenoscopic, line 142: In this case I think we should set `vramSize = 1` and similarly override the memory estimation behavior below to match the previous behavior (i.e. `1` per model), otherwise I don't think we'll be able to load any models.

pkg/inference/scheduling/loader.go
```diff
@@ -132,20 +145,27 @@ func newLoader(
...
+	} else {
+		log.Infof("Running on system with %dMB VRAM", vramSize/1024.0/1024.0)
```
xenoscopic: Curious about the `float` division here.  Does the typechecker allow that?  Would `/1024/1024` not work as expected?

p1-0tr: Must be an artifact of me copying stuff around, this should be integer division.

pkg/inference/scheduling/loader.go
```diff
@@ -132,20 +145,27 @@ func newLoader(
...
+	}
+	hostInfo, err := sysinfo.Host()
+	if err != nil {
+		log.Warnf("Could not read host info: %s", err)
+	}
+	ramSize, err := hostInfo.Memory()
```
doringeman: The above error check should return, otherwise I'm afraid this will panic.

pkg/inference/scheduling/loader.go
```diff
@@ -56,6 +58,14 @@ type runnerInfo struct {
 	modelRef string
 }
 
+// memory is used to keep track of runner memory allocations, and available memory size.
+type memory struct {
```
doringeman: Any strong reason for not using `RequiredMemory`?

pkg/inference/scheduling/loader.go
```diff
@@ -399,15 +421,16 @@ func (l *loader) load(ctx context.Context, backendName, modelID, modelRef string
...
-	// schedule appropriately. Thus, we should switch to computing model size
-	// here through estimation (using parameter count and quantization data type
-	// size).
-	memory := uint64(1)
-	if memory > l.totalMemory {
+	var runnerConfig *inference.BackendConfiguration
+	if rc, ok := l.runnerConfigs[runnerKey{backendName, modelID, mode}]; ok {
+		runnerConfig = &rc
+	}
+	memory, err := backend.GetRequiredMemoryForModel(modelID, runnerConfig)
```
doringeman, line 422: I'd like to store this so I can see it on `docker model ps` (`GetRunningBackends`).

pkg/inference/backends/llamacpp/llamacpp.go
p1-0tr: Yep, I think returning `1` as estimated VRAM requirement on `win/arm64` makes sense. It will still require patching in case other quantisations become supported, though.

pkg/inference/scheduling/loader.go
p1-0tr: Nope, not really.

pkg/inference/scheduling/loader.go
p1-0tr: It does get stored in `allocations`. I can follow up with a PR to add a field in `BackendStatus`

pkg/inference/scheduling/loader.go
```diff
@@ -132,20 +135,31 @@ func newLoader(
...
-	totalMemory := uint64(1)
-	if isGPUEnabledCloudEnvironment {
-		totalMemory = 2
+	// TODO(p1-0tr): improve error handling
+	vramSize, err := gpuInfo.GetVRAMSize()
+	if err != nil {
+		vramSize = 1
+		log.Warnf("Could not read VRAM size: %s", err)
+	} else {
+		log.Infof("Running on system with %dMB VRAM", vramSize/1022/1024)
```
doringeman:
```suggestion
		log.Infof("Running on system with %dMB VRAM", vramSize/1024/1024)
```

pkg/inference/scheduling/loader.go
```diff
@@ -132,20 +135,31 @@ func newLoader(
...
-	// VRAM size here (and potentially even reserving a portion of it) and
-	// computing model size through estimation (using parameter count and
-	// quantization data type size).
-	//
-	// HACK: On GPU-enabled cloud engines, we'll bump this to 2. We can remove
-	// this once we have VRAM estimation.
-	totalMemory := uint64(1)
-	if isGPUEnabledCloudEnvironment {
-		totalMemory = 2
+	// TODO(p1-0tr): improve error handling
```
doringeman, line 138:
```suggestion
```
? Or do you plan something more?

doringeman (approved):
LGTM!

pkg/inference/scheduling/loader.go
p1-0tr: I'm on the fence. On the one hand the current best effort logic is neat in a way. But on the other it would be much cleaner to just fail if we can't detect read the RAM and VRAM sizes (perhaps with a chicken-switch to disable the memory estimation logic altogether). Anyway, it's something for a follow up PR :)

pkg/inference/backends/llamacpp/llamacpp.go
```diff
@@ -213,3 +222,86 @@ func (l *llamaCpp) GetDiskUsage() (int64, error) {
...
+	var vram uint64
+	if len(estimate.Devices) > 1 {
+		vram = uint64(estimate.Devices[1].Weight.Sum() + estimate.Devices[1].KVCache.Sum() + estimate.Devices[1].Computation.Sum())
+	}
+
+	if runtime.GOOS == "windows" && runtime.GOARCH == "arm64" {
+		// TODO(p1-0tr): For now on windows/arm64 stick to the old behaviour, of allowing
+		// one model at a time. This WA requires gpuinfo.GetVRAMSize to return 1.
+		vram = 1
+	}
```
xenoscopic, line 271: nit: could just move this (or equivalent) to top of method to save a bit of overhead.

xenoscopic (approved):
LGTM!  I ran some basic smoke tests on my Mac.  I think we should merge and get it into nightlies ASAP.

PR #111, opened 2025-07-18
recorder: use model ref rather than model ID

xenoscopic:
The model ID is already recorded elsewhere and used as the record key.


p1-0tr (approved):
LGTM

xenoscopic:
A bit more work seems to be needed here, I think both on our side and the Pinata side; I'll finish it up Monday.

PR #113, opened 2025-07-18
Adds model load endpoint to models API

ekcasey:
Add `POST /models/load` endpoint to the model manager which allows TAR formatted models to by streamed from the client and stored in the archive. See https://github.com/docker/model-distribution/pull/113 for formatting details.

Endpoint should be analagous to [`POST images/load`](https://docs.docker.com/reference/api/engine/version/v1.45/#tag/Image/operation/ImageLoad) in the docker engine API.

### DIFF
--- a/pkg/inference/models/manager.go
+++ b/pkg/inference/models/manager.go
@@ -15,10 +15,11 @@ import (
 	"github.com/docker/model-distribution/distribution"
 	"github.com/docker/model-distribution/registry"
 	"github.com/docker/model-distribution/types"
+	"github.com/sirupsen/logrus"
+
 	"github.com/docker/model-runner/pkg/diskusage"
 	"github.com/docker/model-runner/pkg/inference"
 	"github.com/docker/model-runner/pkg/logging"
-	"github.com/sirupsen/logrus"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


p1-0tr (approved):

pkg/inference/models/manager.go
```diff
@@ -187,6 +189,22 @@ func (m *Manager) handleCreateModel(w http.ResponseWriter, r *http.Request) {
...
+// handleLoadModel handles POST <inference-prefix>/models/load requests.
+func (m *Manager) handleLoadModel(w http.ResponseWriter, r *http.Request) {
+	if m.distributionClient == nil {
+		http.Error(w, "model distribution service unavailable", http.StatusServiceUnavailable)
+		return
+	}
+
+	defer r.Body.Close()
+
+	if _, err := m.distributionClient.LoadModel(r.Body, w); err != nil {
```
xenoscopic:
```suggestion
	}

	if _, err := m.distributionClient.LoadModel(r.Body, w); err != nil {
```
For server request objects, the HTTP server [automatically closes](https://pkg.go.dev/net/http#Request) the body:

```
	// The Server will close the request body. The ServeHTTP
	// Handler does not need to.

doringeman: Right...

PR #116, opened 2025-07-22
fix: switch to a RWMutex for synchronizing the router rebuild

doringeman:
Global locks in ServeHTTP methods were preventing concurrent request processing. Individual handlers should manage their own synchronization.

E.g., without this patch, a `model ls` is hanging during a `model pull`.


p1-0tr:
Are we sure there are no behaviours which ended up relying on these locks?

xenoscopic (approved):

PR #123, opened 2025-07-29
charts: add Kubernetes examples

nicks:
- a helm chart
- static Kubernetes configs for a few common setups

I put these under ./charts so we can expose
this as a Helm chart repo later if we want,
but for now we'll just tell people to install it
from source.

Signed-off-by: Nick Santos <nick.santos@docker.com>


xenoscopic (approved):
Looks fantastic - thanks so much @nicks!  I'll let anyone else on the team have a look if they want, but otherwise we can merge tomorrow.

PR #125, opened 2025-07-30
Memory estimation for remote models

p1-0tr:


pkg/inference/models/manager.go
```diff
@@ -163,6 +168,17 @@ func (m *Manager) handleCreateModel(w http.ResponseWriter, r *http.Request) {
 
 	// Pull the model. In the future, we may support additional operations here
 	// besides pulling (such as model building).
+	haveMem, err := m.memoryEstimator.HaveSufficientMemoryForModel(r.Context(), request.From, nil)
+	if err != nil {
+		m.log.Warnf("Failed to calculate memory required for model %q: %s", request.From, err)
+		http.Error(w, "Could not calculate runtime memory requirement for model", http.StatusInternalServerError)
+		return
```
ekcasey: If something goes wrong with the calculation maybe we want to log an error and continue, so that an issue with memory estimation doesn't make the runner unusable?

pkg/inference/models/manager.go
```diff
@@ -163,6 +168,17 @@ func (m *Manager) handleCreateModel(w http.ResponseWriter, r *http.Request) {
...
+	}
+	if !haveMem {
+		m.log.Warnf("Runtime memory requirement for model %q exceeds total system memory", request.From)
+		http.Error(w, "Runtime memory requirement for model exceeds total system memory", http.StatusInsufficientStorage)
+		return
```
ekcasey: I was considering wether there were any cases where a user may want to pull a model on a system where there is insufficient memory to run it.

Some edge cases I thought of:
1. transferring between registries where you might pull, tag, and then push without ever running the model
2. maybe some future packaging use cases where you are pulling a model and then extending/modifying it?? but in these cases I would actually prefer if we accomplish this without the need to pull remote layers

If we think there are any potential uses for pulling a model to a system where it can't run we may want to add a `force` param to explicitly ignore the check.

pkg/inference/models/manager.go
```diff
@@ -571,6 +592,33 @@ func (m *Manager) GetModel(ref string) (types.Model, error) {
...
+}
+
+// BearerTokenForModel returns the bearer token needed to pull a given model.
+func (m *Manager) BearerTokenForModel(ctx context.Context, ref string) (string, error) {
+	tok, err := m.registryClient.BearerToken(ctx, ref)
+	if err != nil {
+		return "", fmt.Errorf("error while getting bearer token for model: %w", err)
+	}
+	return tok, nil
+}
```
ekcasey, line 623: thoughts for the future:

Seems like we are adding a lot of methods to the Manager that are direct pass throughs to distribution or registry client.

In the future maybe we should use the clients from MD directly when a go API is needed. And  leave ModelManager with the single responsibility of exposing this functional via an HTTP API.

pkg/inference/backends/llamacpp/llamacpp.go
```diff
@@ -276,6 +276,71 @@ func (l *llamaCpp) GetRequiredMemoryForModel(model string, config *inference.Bac
 	}, nil
 }
 
+func (l *llamaCpp) parseLocalModel(model string) (*parser.GGUFFile, types.Config, error) {
```
ekcasey, line 282: thoughts for the future: I am on board with doing what is convenient for now, especially when we only have one backend where estimating is really supported. But we may eventually what to move the part of this logic that is tied to the format (GGUF) out of here and into distribution (where we already do some GGUF parsing) and leave only the

ekcasey (commented):
I know this is still in draft but I took as look as I'll be out next week. It looks good to me! But I think it might be safer to proceed in the case of an estimation error and also provide an escape hatch so we don't prevent users who know what they are doing from doing valid things we may not have anticipated.

Also left some thoughts for future refactorings that occurred to me while looking through. Not necessarily suggesting we tackle these now, but I do think some of the interfaces will need a cleanup eventually so I'd thought I'd capture some thoughts as I went

pkg/inference/models/manager.go
p1-0tr: Yup, definitely. I was thinking of putting in a config option to enable/disable resource-related checks.

go.sum
```diff
@@ -40,6 +40,10 @@ github.com/docker/docker-credential-helpers v0.8.2 h1:bX3YxiGzFP5sOXWc3bTPEXdEaZ
 github.com/docker/docker-credential-helpers v0.8.2/go.mod h1:P3ci7E3lwkZg6XiHdRKft1KckHiO9a2rNtyFbZ/ry9M=
 github.com/docker/model-distribution v0.0.0-20250724114133-a11d745e582c h1:w9MekYamXmWLe9ZWXWgNXJ7BLDDemXwB8WcF7wzHF5Q=
 github.com/docker/model-distribution v0.0.0-20250724114133-a11d745e582c/go.mod h1:dThpO9JoG5Px3i+rTluAeZcqLGw8C0qepuEL4gL2o/c=
+github.com/docker/model-distribution v0.0.0-20250730100137-5df88ad2f885 h1:0DSMXzWt3yCjSOVILau11/vrdtQnsHb/OMqNYVU39H0=
+github.com/docker/model-distribution v0.0.0-20250730100137-5df88ad2f885/go.mod h1:dThpO9JoG5Px3i+rTluAeZcqLGw8C0qepuEL4gL2o/c=
+github.com/docker/model-distribution v0.0.0-20250811072316-8ae9665e6889 h1:O1m0yG1N2t6qV8MWT/pe9Z9ukaV0+BX27gg8fsFXDKk=
+github.com/docker/model-distribution v0.0.0-20250811072316-8ae9665e6889/go.mod h1:dThpO9JoG5Px3i+rTluAeZcqLGw8C0qepuEL4gL2o/c=
```
xenoscopic: nit: can we `go mod tidy`?

pkg/inference/memory/system.go
```diff
@@ -0,0 +1,49 @@
...
+		if err != nil {
+			log.Warnf("Could not read host RAM size: %s", err)
+		} else {
+			ramSize = ram.Total
+			log.Infof("Running on system with %dMB RAM", ramSize/1024/1024)
+		}
+	}
+	return &systemMemoryInfo{
+		log:         log,
+		totalMemory: inference.RequiredMemory{RAM: ramSize, VRAM: vramSize}}, nil
```
xenoscopic:
```suggestion
		totalMemory: inference.RequiredMemory{RAM: ramSize, VRAM: vramSize},
	}, nil
```
style nit: it's typically more idiomatic (e.g. in golang/go) in multi-line value construction to use a trailing `,` and move the bracket to the next line.

pkg/inference/memory/system.go
```diff
@@ -0,0 +1,49 @@
...
+
+func NewSystemMemoryInfo(log logging.Logger, gpuInfo *gpuinfo.GPUInfo) (SystemMemoryInfo, error) {
+	// Compute the amount of available memory.
+	// TODO(p1-0tr): improve error handling
+	vramSize, err := gpuInfo.GetVRAMSize()
+	if err != nil {
+		vramSize = 1
+		log.Warnf("Could not read VRAM size: %s", err)
+	} else {
+		log.Infof("Running on system with %dMB VRAM", vramSize/1024/1024)
```
xenoscopic:
```suggestion
		log.Infof("Running on system with %d MB VRAM", vramSize/1024/1024)
```
nit: legibility

pkg/inference/memory/system.go
```diff
@@ -0,0 +1,49 @@
...
+	hostInfo, err := sysinfo.Host()
+	if err != nil {
+		log.Warnf("Could not read host info: %s", err)
+	} else {
+		ram, err := hostInfo.Memory()
+		if err != nil {
+			log.Warnf("Could not read host RAM size: %s", err)
+		} else {
+			ramSize = ram.Total
+			log.Infof("Running on system with %dMB RAM", ramSize/1024/1024)
```
xenoscopic:
```suggestion
			log.Infof("Running on system with %d MB RAM", ramSize/1024/1024)
```
nit: legibility

PR #133, opened 2025-08-22
Support GGUF shards

ekcasey:
* `llama.cpp` backend supports models that are split across multiple GGUF shards.
* Executes from runtime bundle instead of pointing to raw blob paths from content store.
* See https://github.com/docker/model-distribution/pull/123 for details on model and bundle formats

** In draft until https://github.com/docker/model-distribution/pull/123 merges and we can repoint at main branch **

### DIFF
--- a/pkg/inference/backends/llamacpp/llamacpp_config.go
+++ b/pkg/inference/backends/llamacpp/llamacpp_config.go
@@ -62,8 +62,7 @@ func (c *Config) GetArgs(bundle types.ModelBundle, socket string, mode inference
 	}
 
 	// Add arguments for Multimodal projector
-	mmprojPath := bundle.MMPROJPath()
-	if path := mmprojPath; path != "" {
+	if path := bundle.MMPROJPath(); path != "" {
 		args = append(args, "--mmproj", path)
 	}
 
--- a/pkg/inference/backends/llamacpp/llamacpp_config.go
+++ b/pkg/inference/backends/llamacpp/llamacpp_config.go
@@ -6,6 +6,7 @@ import (
 	"strconv"
 
 	"github.com/docker/model-distribution/types"
+
 	"github.com/docker/model-runner/pkg/inference"
 )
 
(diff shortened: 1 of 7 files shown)


pkg/inference/backends/llamacpp/llamacpp_config.go
```diff
@@ -57,25 +53,27 @@ func (c *Config) GetArgs(model types.Model, socket string, mode inference.Backen
...
 	// Add arguments from backend config
 	if config != nil {
 		args = append(args, config.RuntimeFlags...)
 	}
 
 	// Add arguments for Multimodal projector
-	path, err := model.MMPROJPath()
-	if path != "" && err == nil {
+	mmprojPath := bundle.MMPROJPath()
+	fmt.Println("GOT MMPROJ:", mmprojPath)
```
p1-0tr: nit: looks like a leftover debug print. Could be useful as an `c.log.Info()`, though the runtime args should be logged somewhere around the point of execution.

ekcasey: oops! thank you

p1-0tr (approved):
LGTM

pkg/inference/backends/llamacpp/llamacpp_config.go
```diff
@@ -57,25 +53,26 @@ func (c *Config) GetArgs(model types.Model, socket string, mode inference.Backen
...
 		args = append(args, config.RuntimeFlags...)
 	}
 
 	// Add arguments for Multimodal projector
-	path, err := model.MMPROJPath()
-	if path != "" && err == nil {
+	mmprojPath := bundle.MMPROJPath()
+	if path := mmprojPath; path != "" {
 		args = append(args, "--mmproj", path)
 	}
```
xenoscopic:
```suggestion
	if path := bundle.MMPROJPath(); path != "" {
		args = append(args, "--mmproj", path)
	}
```
Doesn't seem like `mmprojPath` is used again?

xenoscopic (approved):
Looks good, I think just one code path I'd clean up slightly.

PR #137, opened 2025-08-22
Fix remote memory estimation

ekcasey:
* pull in blob URL fix - https://github.com/docker/model-distribution/pull/123
* don't attempt to estimate sharded models

### DIFF
--- a/pkg/inference/backends/llamacpp/llamacpp.go
+++ b/pkg/inference/backends/llamacpp/llamacpp.go
@@ -19,6 +19,7 @@ import (
 	parser "github.com/gpustack/gguf-parser-go"
 
 	"github.com/docker/model-distribution/types"
+
 	"github.com/docker/model-runner/pkg/diskusage"
 	"github.com/docker/model-runner/pkg/inference"
 	"github.com/docker/model-runner/pkg/inference/config"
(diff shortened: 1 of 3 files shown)


pkg/inference/backends/llamacpp/llamacpp.go
```diff
@@ -308,19 +309,15 @@ func (l *llamaCpp) parseRemoteModel(ctx context.Context, model string) (*parser.
...
-		}
-		if mt == types.MediaTypeGGUF {
-			ggufDigest, err = layer.Digest()
-			if err != nil {
-				return nil, types.Config{}, fmt.Errorf("getting digest of GGUF layer for model(%s): %w", model, err)
-			}
-			break
-		}
+	ggufLayers := getGGUFLayers(layers)
+	if len(ggufLayers) != 0 {
```
xenoscopic:
```suggestion
	if len(ggufLayers) != 1 {
```

ekcasey: just fixing that lol

xenoscopic (approved):
Looks good to me, but @p1-0tr's +1 is probably more substantial here.

PR #144, opened 2025-08-25
fix: move CORS middleware to top level to handle preflight requests

fiam:
Previously, CORS preflight requests were not working because the CORS
middleware was nested within the inference package. Moving it to a
dedicated middleware package at the top level ensures preflight
requests are properly handled before reaching route handlers.

- Move CORS implementation from pkg/inference to pkg/middleware
- Add comprehensive test coverage for CORS middleware
- Add unit tests for inference models manager and scheduler

Signed-off-by: Alberto Garcia Hierro <damaso.hierro@docker.com>

### DIFF
--- a/pkg/inference/models/manager.go
+++ b/pkg/inference/models/manager.go
@@ -19,6 +19,7 @@ import (
 	"github.com/docker/model-runner/pkg/inference"
 	"github.com/docker/model-runner/pkg/inference/memory"
 	"github.com/docker/model-runner/pkg/logging"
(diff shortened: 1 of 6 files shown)


fiam:
References I used:

- https://en.wikipedia.org/wiki/Cross-origin_resource_sharing
- https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request
- https://developer.mozilla.org/en-US/docs/Glossary/CORS

doringeman (approved):
LGTM! Thanks!

PR #147, opened 2025-08-28
refactor(OpenAIRecorder): use Unix timestamp instead of time.Time

doringeman:
Use Unix timestamp instead of `time.Time` to ease the sorting.


p1-0tr (approved):
LGTM

PR #148, opened 2025-08-29
inference: Return memory requirement in estimation error

p1-0tr:


pkg/inference/memory/system.go
```diff
@@ -40,6 +40,8 @@ func NewSystemMemoryInfo(log logging.Logger, gpuInfo *gpuinfo.GPUInfo) (SystemMe
 			log.Infof("Running on system with %d MB RAM", ramSize/1024/1024)
 		}
 	}
+	ramSize = 1024 * 1024 * 1024
+	vramSize = 1024 * 1024 * 1024
```
xenoscopic: Was this left in during debugging?

pkg/inference/memory/estimator.go
```diff
@@ -39,10 +39,11 @@ func (m *memoryEstimator) GetRequiredMemoryForModel(ctx context.Context, model s
 	return m.defaultBackend.GetRequiredMemoryForModel(ctx, model, config)
 }
 
-func (m *memoryEstimator) HaveSufficientMemoryForModel(ctx context.Context, model string, config *inference.BackendConfiguration) (bool, error) {
+func (m *memoryEstimator) HaveSufficientMemoryForModel(ctx context.Context, model string, config *inference.BackendConfiguration) (bool, *inference.RequiredMemory, *inference.RequiredMemory, error) {
```
xenoscopic: Should we try to unify pointer vs. non-pointer return types?  `inference.RequiredMemory` seems it should always be passed / returned as a non-pointer type since it's small (as it is in `SystemMemoryInfo`'s methods).

pkg/inference/memory/system.go
p1-0tr: yep 🤦

pkg/inference/memory/estimator.go
p1-0tr: done :)

xenoscopic (approved):
LGTM!

PR #149, opened 2025-08-29
Implement llama.cpp sandboxing for macOS and Windows

xenoscopic:
This PR implements llama.cpp process sandboxing for macOS and Windows.  It defines an extensible sandboxing framework that we can use for other backends or critical processes.

On macOS, we use the `sandbox-exec` command with a strict profile (similar to [what Chromium does](https://www.chromium.org/developers/design-documents/sandbox/osx-sandboxing-design/)). This command is deprecated, but the only other option would be to use App Sandbox, which we can't do for a standalone process. In theory, we may inherit something of a sandbox from Docker Desktop (if not running standalone), but Docker Desktop's entitlements are *broad*, so they won't do much to contain a compromised llama.cpp server process.  We can refine this profile to be more aggressive over time by applying more granular system call restrictions, but this should already significantly boost security.  Unfortunately this API is deprecated, so defining sandbox profiles isn't documented (it uses a custom LISP dialect). The best option is to look in `/System/Library/Sandbox/Profiles` or Google a bit. (**Edit:** it looks like Chromium preserved the [documentation](http://reverse.put.as/wp-content/uploads/2011/09/Apple-Sandbox-Guide-v1.0.pdf)).

On Windows, we use the Job API (similar to [what Chromium does](https://chromium.googlesource.com/chromium/src/+/refs/heads/lkgr/docs/design/sandbox.md#The-Job-object)).  There is also a [Windows Sandbox](https://learn.microsoft.com/en-us/windows/security/application-security/application-isolation/windows-sandbox/) mechanism available, but it isn't available by default and it uses virtualization.  It might be worth looking at in the future.  In the meantime, we can use a restricted Windows Job object to provide at least some encapsulation.  I've built a little DSL / parser for specifying configuration there, sort of in the spirit of the macOS sandbox profile DSL.

On Linux, we just pass through directly to `os/exec` (i.e. we don't currently use sandboxing).  There isn't a unified sandboxing API on Linux (ideally we'd want something like OpenBSD's `tame()`), so it's not easy to choose something.  However, we don't support Docker Model Runner on Docker Desktop for Linux, and on Docker CE / Offload, DMR runs inside a container, so there's not much need for additional sandboxing.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -60,7 +60,7 @@ require (
 	golang.org/x/crypto v0.37.0 // indirect
 	golang.org/x/exp v0.0.0-20250106191152-7588d65b2ba8 // indirect
 	golang.org/x/mod v0.24.0 // indirect
-	golang.org/x/sys v0.33.0 // indirect
+	golang.org/x/sys v0.35.0 // indirect
 	golang.org/x/tools v0.32.0 // indirect
 	gonum.org/v1/gonum v0.15.1 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20250414145226-207652e42e2e // indirect
@@ -69,3 +69,5 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	howett.net/plist v1.0.0 // indirect
 )
+
+replace github.com/kolesnikovae/go-winjob => github.com/docker/go-winjob v0.0.0-20250829235554-57b487ebcbc5
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/google/go-containerregistry v0.20.3
 	github.com/gpustack/gguf-parser-go v0.14.1
 	github.com/jaypipes/ghw v0.16.0
+	github.com/kolesnikovae/go-winjob v1.0.0
 	github.com/mattn/go-shellwords v1.0.12
 	github.com/opencontainers/go-digest v1.0.0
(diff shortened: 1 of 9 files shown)


xenoscopic:
This is now ready for review.  I've tested it on macOS (`arm64`) and Windows (`amd64` and `arm64` (the latter both with and without OpenCL enabled))).

**The one thing that needs testing is Windows on amd64 with CUDA enabled**.

Docker Desktop test builds here:

https://desktop-stage.docker.com/win/main/arm64/203302/Docker%20Desktop%20Installer.exe
https://desktop-stage.docker.com/win/main/amd64/203302/Docker%20Desktop%20Installer.exe
https://desktop-stage.docker.com/mac/main/arm64/203302/Docker.dmg

You can also test in standalone mode.

Note that when we update Docker Desktop, we'll need to replicate the `replace` directive here that's in `go.mod`:

```
github.com/kolesnikovae/go-winjob => github.com/docker/go-winjob v0.0.0-20250829235554-57b487ebcbc5
```

I had to patch the `go-winjob` library to support arm64.

xenoscopic:
Here's the what sandboxed llama.cpp process looks like on Windows:

windows_sandbox

macOS doesn't have an equivalent way to view the sandbox profile on a process, but you can play with the profile in the code to verify that it applies (e.g. try removing an `allow` statement).

doringeman (approved):
Nice!

PR #151, opened 2025-09-03
Add colors to logs if suitable

ericcurtin:
Makes it easier to read the log levels of logs.

### DIFF
--- a/pkg/inference/backends/llamacpp/llamacpp_config.go
+++ b/pkg/inference/backends/llamacpp/llamacpp_config.go
@@ -2,6 +2,7 @@ package llamacpp
 
 import (
 	"fmt"
+	"os"
 	"runtime"
 	"strconv"
 
@@ -16,9 +17,34 @@ type Config struct {
 	Args []string
 }
 
+// isTerminal returns true if the given file is a character device (a terminal).
+func isTerminal(f *os.File) bool {
+	fi, err := f.Stat()
+	if err != nil {
+		return false
+	}
+
+	return (fi.Mode() & os.ModeCharDevice) != 0
+}
+
+func stdoutIsTerminal() bool { return isTerminal(os.Stdout) }
+func stderrIsTerminal() bool { return isTerminal(os.Stderr) }
+
+func shouldColorize() bool {
+	term := os.Getenv("TERM")
+	if term == "" || term == "dumb" {
+		return false
+	}
+
+	return stdoutIsTerminal() && stderrIsTerminal()
+}
+
 // NewDefaultLlamaCppConfig creates a new LlamaCppConfig with default values.
 func NewDefaultLlamaCppConfig() *Config {
 	args := append([]string{"--jinja", "-ngl", "100", "--metrics"})
+	if shouldColorize() {
+		args = append(args, "--log-colors")
+	}
 
 	// Special case for Windows ARM64
 	if runtime.GOOS == "windows" && runtime.GOARCH == "arm64" {


pkg/inference/backends/llamacpp/llamacpp_config.go
```diff
@@ -16,9 +17,26 @@ type Config struct {
 	Args []string
 }
 
+func shouldColorize() bool {
+	term := os.Getenv("TERM")
+	if term == "" || term == "dumb" {
+		return false
+	}
+
+	fi, err := os.Stdout.Stat()
```
p1-0tr: I don't think we can safely assume that the writer which the logs will flow through will be underlied by stdout. In fact in most cases I'd assume the logs will flow into a file, which we would not want console control characters in.

pkg/inference/backends/llamacpp/llamacpp_config.go
```diff
@@ -16,9 +17,26 @@ type Config struct {
...
+		return false
+	}
+
+	return (fi.Mode() & os.ModeCharDevice) != 0
+}
+
 // NewDefaultLlamaCppConfig creates a new LlamaCppConfig with default values.
 func NewDefaultLlamaCppConfig() *Config {
 	args := append([]string{"--jinja", "-ngl", "100", "--metrics"})
+	if shouldColorize() {
```
p1-0tr, line 45: TBH this feels like a feature that ideally should be implemented in `model-cli` (though I don't see a good way of doing it)

pkg/inference/backends/llamacpp/llamacpp_config.go
ericcurtin: That's what this code does, we only color if there is tty stdout (outputting to file is one of those cases there is no tty stdout, that's how we flow into a file), right kind of term, etc. This is the same logic battle tested tools like systemd use when deciding when to use colours or not. It's why those kind of tools can have colors on by default (if appropriate, off in all the cases that aren't appropriate). grep would be another example. Seems to work fine for those.

p1-0tr: This code checks if the stdout of the `model-runner` process is a char device (e.g. a tty), which indeed is a sensible condition to use when deciding to output colours or not. Not arguing with that. My point is that the logger into which the output of `llama-server` is forwarded, does not need to be stdout or a stream which eventually flows into stdout. In fact, here - https://github.com/docker/model-runner/blob/main/pkg/inference/backends/llamacpp/llamacpp.go#L160 - stderr of the `llama-server` process is captured so that it can be included in error strings if necessary, apart from being forwarded to the logger.

ericcurtin: I created a `func stdoutIsTerminal() bool` function to make this more clear

ericcurtin: Added check for both stderr and stdout to solve that condition

ericcurtin: stderr is the better check as that's where the logs go...

p1-0tr: That still does not address my key concern. The output  of `llama-server` does not need to be written to stdout or stderr. Where it goes is defined here - https://github.com/docker/model-runner/blob/main/pkg/inference/backends/llamacpp/llamacpp.go#L171 - to make the checks correct they would have to test whether `out` and `serverLogStream` are underlied by a char device, rather than whether the standard streams of the parent process.

pkg/inference/backends/llamacpp/llamacpp_config.go
ericcurtin: This PR aside. We should have ways of passing command line args from model-cli to places like llama-server. There are going to be cases where users will want to manually override certain defaults, etc. Maybe they are fine-tuning something, but there are many reasons. We can set reasonable defaults, but it won't be suitable for everything and everyone.

pkg/inference/backends/llamacpp/llamacpp_config.go
```diff
@@ -16,9 +17,34 @@ type Config struct {
...
+func stdoutIsTerminal() bool { return isTerminal(os.Stdout) }
+func stderrIsTerminal() bool { return isTerminal(os.Stderr) }
+
+func shouldColorize() bool {
+	term := os.Getenv("TERM")
+	if term == "" || term == "dumb" {
+		return false
+	}
+
+	return stdoutIsTerminal() && stderrIsTerminal()
```
ericcurtin, line 39: Maybe it's better we just check stderr. WDYT @p1-0tr ?

pkg/inference/backends/llamacpp/llamacpp_config.go
p1-0tr: Yup, definitely. The UX for this is not quite sorted out yet. There is an early implementation of config support here - https://github.com/docker/model-runner/blob/main/pkg/inference/scheduling/scheduler.go#L366

pkg/inference/backends/llamacpp/llamacpp_config.go
p1-0tr: We need to check where `log.Writer()` goes, because - https://github.com/docker/model-runner/blob/main/pkg/inference/backends/llamacpp/llamacpp.go#L159.

Also we'd need to decorate `tailBuf`, here - https://github.com/docker/model-runner/blob/main/pkg/inference/backends/llamacpp/llamacpp.go#L160, to strip console control characters. Otherwise we'll end up sticking them in errors we return.

PR #152, opened 2025-09-04
Set ngl to 999 by default to align with upstream

ericcurtin:
Upstream now defaults to 999 which means use max available layer.

We could try and rely on the upstream defaults, but there is one case where we manually set to 0 which makes it tricky so maybe aligning is better for now.


xenoscopic:
Adding @p1-0tr as reviewer; he'll know the correct timing to merge based on our llama.cpp backend updates.

p1-0tr (approved):
LGTM

ericcurtin:
Another interesting flag is this one:

```
-c,    --ctx-size N                     size of the prompt context (default: 4096, 0 = loaded from model)
                                        (env: LLAMA_ARG_CTX_SIZE)
```

I think upstream most commonly recommends 0 these days, for example huggingface docs recommend that. But some people set it as 2048 or 4096 because the value in the gguf is set badly.

One I think that should be set also is `--cache-reuse 256` the additional memory required is ~1% but the optimisation benefits are significant. I'm not sure if @JohannesGaessler and @slaren intend on changing the --cache-reuse default upstream from 0 to some non-zero value.

JohannesGaessler:
I have no intention of touching `--cache-reuse` in the foreseeable future.

ericcurtin:
> I have no intention of touching `--cache-reuse` in the foreseeable future.

Because you believe it's not worth it or just don't care enough? (Just curious of your opinions on it)

Would you recommend people use 256 for example? :smile: 

I know @ggerganov has recommended it in the past, he would have done more widespread benchmarks than me.

JohannesGaessler:
I am not very knowledgeable when it comes to the code for KV cache management so I don't think it's worth the opportunity cost for me to work on anything like that. More generally, I think nowadays the context size of most models is large enough that context shifting is not as important.

ericcurtin:
> I am not very knowledgeable when it comes to the code for KV cache management so I don't think it's worth the opportunity cost for me to work on anything like that. More generally, I think nowadays the context size of most models is large enough that context shifting is not as important.

So reading this, your belief is --ctx-size 0 is the best default today

JohannesGaessler:
I think that right now all default values have downsides, `-c 0` will frequently OOM. I'm currently going back to automatically fitting runtime parameters to VRAM, primarily for which tensors to run on GPU but the context size could also be set automatically.

PR #160, opened 2025-09-15
Fix request/response recording 

ilopezluna:
Capture errors on streaming requests and ensure to return a valid json as an error

### DIFF
--- a/pkg/metrics/openai_recorder.go
+++ b/pkg/metrics/openai_recorder.go
@@ -214,6 +214,39 @@ func (r *OpenAIRecorder) normalizeErrorToJSON(errorContent string) string {
 	return fmt.Sprintf(`{"error": %s}`, string(escapedContent))
 }
 
+// handleErrorRecording handles the logic for recording errors and responses based on
+// streaming errors and HTTP status codes.
+func (r *OpenAIRecorder) handleErrorRecording(record *RequestResponsePair, streamingErr error, response string, statusCode int) {
+	if streamingErr != nil {
+		record.Error = r.serializeStreamingError(streamingErr)
+		record.Response = ""
+		return
+	}
+
+	if statusCode >= 400 {
+		record.Error = r.normalizeErrorToJSON(response)
+		record.Response = ""
+		return
+	}
+
+	// Success case
+	record.Response = response
+	record.Error = ""
+}
+
+// serializeStreamingError handles the serialization of streaming errors.
+// It attempts to serialize StreamingError types directly to JSON for rich error structure,
+// falling back to normalized JSON for other error types.
+func (r *OpenAIRecorder) serializeStreamingError(err error) string {
+	var streamingError *StreamingError
+	if errors.As(err, &streamingError) {
+		if errorJSON, marshalErr := json.Marshal(streamingError); marshalErr == nil {
+			return string(errorJSON)
+		}
+	}
+	return r.normalizeErrorToJSON(err.Error())
+}
+
 func (r *OpenAIRecorder) RecordResponse(id, model string, rw http.ResponseWriter) {
 	rr := rw.(*responseRecorder)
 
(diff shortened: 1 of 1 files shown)


pkg/metrics/openai_recorder.go
```diff
@@ -197,13 +228,31 @@ func (r *OpenAIRecorder) RecordResponse(id, model string, rw http.ResponseWriter
...
 
 	for _, line := range lines {
+		// Check for error lines in streaming format
+		if strings.HasPrefix(line, "error: ") {
+			errorData := strings.TrimPrefix(line, "error: ")
+			var errorObj map[string]interface{}
+			if err := json.Unmarshal([]byte(errorData), &errorObj); err == nil {
+				// Extract error code if available
+				if code, ok := errorObj["code"].(float64); ok {
+					errorStatusCode = int(code)
```
doringeman: Do you know that this can't be 400?
Why don't we return an error instead of an int to signal an error?

ilopezluna: > Do you know that this can't be 400?

Do you mean why the status code is not 400 in this case? If so, it is because in streaming responses the status code is 200 and then it starts streaming chunks of data. In case of error (like the one we can get if we exceed the context-length) the endpoint returns a chunk of data prefixed by `"error: "` 

> Why don't we return an error instead of an int to signal an error?

I think returning an error here is a valid alternative, and probably more go idiomatic, I think we only need to make sure we get include the `errorStatusCode` in the error as well as the `streamingBody`.
I'll update it 👍

ilopezluna: @doringeman  addressed in [26172b0](https://github.com/docker/model-runner/pull/160/commits/26172b04c8f50ada775710863c34c4bf5eb1acd1)

pkg/metrics/openai_recorder.go
```diff
@@ -165,8 +214,9 @@ func (r *OpenAIRecorder) RecordResponse(id, model string, rw http.ResponseWriter
 	}
 
 	var response string
+	var streamingErr error
 	if strings.Contains(responseBody, "data: ") {
-		response = r.convertStreamingResponse(responseBody)
+		response, streamingErr = r.convertStreamingResponse(responseBody)
 	} else {
 		response = responseBody
```
doringeman, line 265: I think we should do the same for non-streamed responses - they might also be non-JSON formatted, right?

ilopezluna: Yeah, `normalizeErrorToJSON(response)` is also called when `if statusCode >= 400 {`
So in case of an error on non-streamed responses we will also ensure we return a json as response

pkg/metrics/openai_recorder.go
```diff
@@ -21,6 +22,31 @@ const maximumRecordsPerModel = 10
 // subscriberChannelBuffer is the buffer size for subscriber channels.
 const subscriberChannelBuffer = 100
 
+// defaultStreamingErrorCode is the default code for streaming errors.
+const defaultStreamingErrorCode = 400
```
xenoscopic:
```suggestion
const defaultStreamingErrorCode = http. StatusBadRequest
```

pkg/metrics/openai_recorder.go
```diff
@@ -162,6 +188,65 @@ func (r *OpenAIRecorder) NewResponseRecorder(w http.ResponseWriter) http.Respons
...
+// handleErrorRecording handles the logic for recording errors and responses based on
+// streaming errors and HTTP status codes.
+func (r *OpenAIRecorder) handleErrorRecording(record *RequestResponsePair, streamingErr error, response string, statusCode int) {
+	if streamingErr != nil {
+		record.Error = r.serializeStreamingError(streamingErr)
+		record.Response = ""
+		return
+	}
+
+	if statusCode >= 400 {
```
xenoscopic, line 226:
```suggestion
	if statusCode >= http.StatusBadRequest {
```

xenoscopic (approved):

doringeman (approved):

PR #165, opened 2025-09-23
Consolidate model-distribution into model-runner monorepo

doringeman:
Bring in https://github.com/docker/model-distribution.
The goal is to centralize all Docker Model Runner components into a single repository, simplifying development, dependency management, and deployment workflows.

## Summary by Sourcery

Consolidate the external model-distribution repository into the model-runner monorepo under pkg/distribution, update imports accordingly, and streamline development and deployment workflows.

New Features:
- Import model-distribution code directly into pkg/distribution and registry subpackages
- Introduce a new CLI tool (cmd/mdltool) for packaging, pushing, pulling, and managing models with cobra

Enhancements:
- Remove external github.com/docker/model-distribution dependency and update go.mod with new local paths and additional dependencies
- Merge transport libraries (resumable, parallel) and internal testing utilities into the monorepo
- Add a top-level Makefile to build and tidy the monorepo

CI:
- Extend CI workflow to build and test the distribution code via make in pkg/distribution

Documentation:
- Update README to reference the local pkg/distribution path instead of the external repo

Tests:
- Integrate and adapt a comprehensive suite of distribution, transport, registry, store, and CLI tests


tools/benchmarks/parallelget/main.go
```diff
@@ -0,0 +1,348 @@
+package main
```
xenoscopic, line 1: It might make more sense to move this directory up to the top level of the repo; it's uncommon to have `main` packages situated anywhere under the top-level `pkg` except in very specific cases like test artifacts.

doringeman:
> I was curious why the PR only has 59 commits vs. github.com/docker/model-distribution's 108 [...]

There are a couple of reasons:
1. We didn't include the patches for these paths: .github, .gitignore, go.mod + go.sum.
```
$ git log --oneline -- .gitignore .github/ go.mod go.sum | wc -l
      35
```
2. Running `git format-patch` for specific paths (i.e., `git format-patch -o /tmp/mdltool-patches --root --numbered --no-renames -- cmd/mdltool`) squashes such changes https://github.com/docker/model-distribution/pull/32/files into https://github.com/docker/model-runner/pull/165/commits/3414ee85b75308a1adc6ca01d49d7e9bc4fe915c only.

> I'm also a little weary of migrating github.com/docker/model-distribution into the same Go module as github.com/docker/model-runner, as I feel it's going to break down the API boundary between them.

My main thinking was: do people use model-distribution independently so it'd be beneficial of having a module for it? But we can revisit this!

> [...] I won't be as active in the codebase anymore.

😢

tools/benchmarks/parallelget/main.go
doringeman: Fair! Done in https://github.com/docker/model-runner/pull/165/commits/4d1d9336061cc53201cbc96168e13708eadf0ac4.

ericcurtin:
This is great work, are we ok to archive https://github.com/docker/model-distribution now leaving a notice that development will be here now?

PR #167, opened 2025-09-24
This also applies to Linux aarch64

ericcurtin:
It's a decent heuristic in general for aarch64.


ericcurtin:
Tested on ampere system:

https://github.com/containers/ramalama/issues/934

I'm 90% sure GOARCH arm64 is aarch64 on Linux. Testing...

pkg/inference/backends/llamacpp/llamacpp_config.go
```diff
@@ -20,8 +20,8 @@ type Config struct {
 func NewDefaultLlamaCppConfig() *Config {
 	args := append([]string{"--jinja", "-ngl", "999", "--metrics"})
 
-	// Special case for Windows ARM64
-	if runtime.GOOS == "windows" && runtime.GOARCH == "arm64" {
+	// Special case for ARM64 (aarch64 on Linux)
+	if runtime.GOARCH == "aarch64" || runtime.GOARCH == "arm64" {
```
doringeman: Moreover, this would be `true` on macOS as well.

ericcurtin:
> I see it's `arm64`.
> 
> ```
> $ docker run --rm -it docker.io/library/golang:1.24.7 sh -c 'arch && go env GOARCH'
> aarch64
> arm64
> ```

Cool, I can delete

pkg/inference/backends/llamacpp/llamacpp_config.go
ericcurtin: Yup, which is actually fine. Threads is irrelevant on macOS as we use the GPU for inferencing.

ericcurtin:
Should be good @doringeman removed aarch64

doringeman (approved):
Also, we don't want

PR #177, opened 2025-09-25
feat: add media field truncation to reduce memory usage

ilopezluna:
OpenAI requests containing base64-encoded images or audio consume excessive memory when stored by the recorder.
This PR fixes that by:
- Truncate `image_url.url` and `input_audio.data` fields to 100 characters
- Preserve request structure and add truncation markers showing bytes removed
- Maintain JSON validity and backward compatibility

```
Before: "data:image/jpeg;base64,/9j/4AAQSkZJRgABAQAAAQABAAD/2wCE..." (2000+ chars)
After:  "data:image/jpeg;base64,/9j/4AAQSkZJRgABAQAAAQABAAD/2wCE...[truncated 1900 chars]"

```


ericcurtin:
LGTM, I think the bot comments are worth addressing

PR #190, opened 2025-10-02
Consolidate model-cli into model-runner monorepo

doringeman:
Compared to https://github.com/docker/model-runner/pull/165#issuecomment-3327642388, this time I used `git filter-repo`.

Consolidated [model-cli](https://github.com/docker/model-cli/tree/86abe2e55407090b7128d1baa526bb369db8901e) (367 commits) + https://github.com/docker/model-runner/commit/b1f0a10b165ecfbf10cedab2ef13b5fba8fcf220 + https://github.com/docker/model-runner/commit/a39f8ce14e1dc5f94db7a6104a3cf5afeb70bacf +https://github.com/docker/model-runner/pull/190/commits/344ca1bc52784f44e834e511880a099e3b836055 + https://github.com/docker/model-runner/pull/190/commits/bbb7e9b4ed49941c95c843338ea19d62caf05eed + https://github.com/docker/model-runner/pull/190/commits/13d99c49e5eda25ab40ed012553bc1c564eff5ac = 372 commits.


p1-0tr:
Hmm, I wonder if stripping `vendor/*` out wouldn't be a good idea

p1-0tr (approved):

.dockerignore
```diff
@@ -1,5 +1,4 @@
 # Version control
-.git/
```
ilopezluna, line 2: Why don't we ignore `.git`?

cmd/cli/.gitignore
```diff
@@ -0,0 +1,3 @@
+model-cli
```
ilopezluna, line 1: we should remove this file and move the content to the root `.gitignore`, right?

cmd/cli/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
ilopezluna: I think we should just use one single license file, it can be done in follow up PR

.dockerignore
doringeman: I need it for `validate-docs`.
```
docker buildx bake validate-docs -f cmd/cli/docker-bake.hcl
```

cmd/cli/.gitignore
doringeman: It works from here as well. I was planning to move it in a follow-up PR along with the Makefile rules.

cmd/cli/LICENSE
doringeman: Done in https://github.com/docker/model-runner/pull/190/commits/e23bc1e876687419799fae55e79f0992be4c4ab8.

PR #192, opened 2025-10-02
feat: add error handling for unsupported model formats in manager

ilopezluna:
Before:
```
MODEL_RUNNER_HOST=http://localhost:13434 docker model pull aistaging/smollm2-vllm
Failed to pull model: pulling aistaging/smollm2-vllm failed with status 500 Internal Server Error: error while pulling model: safetensors models are not currently supported - this runner only supports GGUF format models
```

After:
```
MODEL_RUNNER_HOST=http://localhost:13434 docker model pull aistaging/smollm2-vllm
Failed to pull model: pulling aistaging/smollm2-vllm failed with status 400 Bad Request: safetensors models are not currently supported - this runner only supports GGUF format models
```

### DIFF
--- a/pkg/inference/models/manager.go
+++ b/pkg/inference/models/manager.go
@@ -12,10 +12,10 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/docker/model-runner/pkg/diskusage"
 	"github.com/docker/model-runner/pkg/distribution/distribution"
 	"github.com/docker/model-runner/pkg/distribution/registry"
 	"github.com/docker/model-runner/pkg/distribution/types"
-	"github.com/docker/model-runner/pkg/diskusage"
 	"github.com/docker/model-runner/pkg/inference"
 	"github.com/docker/model-runner/pkg/inference/memory"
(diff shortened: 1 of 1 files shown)


pkg/inference/models/manager.go
```diff
@@ -196,6 +196,11 @@ func (m *Manager) handleCreateModel(w http.ResponseWriter, r *http.Request) {
 			http.Error(w, "Model not found", http.StatusNotFound)
 			return
 		}
+		if errors.Is(err, distribution.ErrUnsupportedFormat) {
+			m.log.Warnf("Unsupported model format for %q: %v", request.From, err)
+			http.Error(w, distribution.ErrUnsupportedFormat.Error(), http.StatusBadRequest)
```
doringeman: Just in case you consider it's more suitable, otherwise LGTM!
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/415
```suggestion
			http.Error(w, distribution.ErrUnsupportedFormat.Error(), http.StatusUnsupportedMediaType)
```

p1-0tr (approved):

PR #204, opened 2025-10-09
Use lunar vulkan

ericcurtin:
I think we have a mismatch here

## Summary by Sourcery

Switch from installing the system libvulkan1 package to downloading and installing the LunarG Vulkan SDK in the setup script for generic and CPU variants, and expose the SDK via environment variables in the Docker image

Enhancements:
- Use LunarG Vulkan SDK (version 1.4.321.1) instead of system libvulkan1 in the apt-install script
- Configure Vulkan SDK environment variables in the Dockerfile

### DIFF
--- a/scripts/apt-install.sh
+++ b/scripts/apt-install.sh
@@ -5,8 +5,16 @@ main() {
 
   apt-get update
   local packages=("ca-certificates")
-  if [ "${LLAMA_SERVER_VARIANT}" = "generic" ] || [ "${LLAMA_SERVER_VARIANT}" = "cpu" ]; then
-      packages+=("libvulkan1")
+  if [ "$LLAMA_SERVER_VARIANT" = "generic" ] || [ "$LLAMA_SERVER_VARIANT" = "cpu" ]; then
+    # Install Vulkan SDK
+    local vulkan_version=1.4.321.1
+    local arch
+    arch=$(uname -m)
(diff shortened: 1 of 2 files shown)


scripts/apt-install.sh
```diff
@@ -5,8 +5,16 @@ main() {
...
   apt-get update
   local packages=("ca-certificates")
-  if [ "${LLAMA_SERVER_VARIANT}" = "generic" ] || [ "${LLAMA_SERVER_VARIANT}" = "cpu" ]; then
-      packages+=("libvulkan1")
+  if [ "$LLAMA_SERVER_VARIANT" = "generic" ] || [ "$LLAMA_SERVER_VARIANT" = "cpu" ]; then
+    # Install Vulkan SDK
+    local vulkan_version=1.4.321.1
+    local arch
+    arch=$(uname -m)
+    wget -qO /tmp/vulkan-sdk.tar.xz https://sdk.lunarg.com/sdk/download/$vulkan_version/linux/vulkan-sdk-linux-"$arch"-$vulkan_version.tar.xz
```
p1-0tr, line 13: Lunar only provide prebuilt packages for `x86_64`, other platforms need to be built from source (see https://github.com/docker/inference-engine-llama.cpp/pull/71/files#diff-041e3c30cdebf29d1bd189bbad5562d3c274c8dbda07c6ff9cbe8cd422a684edR20)

ericcurtin: I think we should just use Ubuntu repositories for ARM... I think Lunar are discontinuing building these packages anyway

PR #205, opened 2025-10-09
Fix vulkan SDK install on linux/arm64

p1-0tr:
## Summary by Sourcery

Fix Vulkan SDK installation on linux/arm64 by installing missing tools, adjusting extraction steps, and building the SDK on non-x86_64 architectures

Bug Fixes:
- Add missing wget and xz-utils dependencies to support SDK download and extraction
- Ensure correct setup of Vulkan SDK on ARM by installing build dependencies and compiling the SDK when arch is not x86_64

Enhancements:
- Adjust tar extraction to preserve versioned directory before moving artifacts to /opt/vulkan
- Clean up redundant TARGETARCH argument in Dockerfile


ericcurtin:
Does this match with how we build in the other repo?

p1-0tr:
> Does this match with how we build in the other repo?

Yep, the only difference is here I went with building the full SDK for now. The runtime repo only builds the components that it needs.

PR #212, opened 2025-10-10
Document simple installation techniques

ericcurtin:
For new users

## Summary by Sourcery

Introduce simple installation instructions for Docker Desktop and Docker Engine in the README to guide new users through setup

Documentation:
- Add an Installation section to the README with instructions for Docker Desktop and Docker Engine
- Include links to Docker Desktop packages and the Docker Engine installer script
- Reference the official Model Runner getting started documentation for more details

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,24 @@ Docker Model Runner (DMR) makes it easy to manage, run, and deploy AI models usi
 
 This package supports the Docker Model Runner in Docker Desktop and Docker Engine.
 
+### Installation
+
+For macOS and Windows, install Docker Desktop:
+
+https://docs.docker.com/desktop/
+
+For Linux, install Docker Engine:
+
+```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,6 +6,22 @@ Docker Model Runner (DMR) makes it easy to manage, run, and deploy AI models usi
 
 This package supports the Docker Model Runner in Docker Desktop and Docker Engine.
 
+### Installation
+
+For Docker Desktop there are packages for macOS, Windows and Linux:
```
p1-0tr: We don't support DMR on Linux DD

ericcurtin: Good catch! I wonder should we change that

p1-0tr: TBH it'd likely create more issues than it'd solve. If we had a lot of users interested in the feature it may be a different question.

PR #213, opened 2025-10-10
standalone: Allow specifying controller image variant via env

p1-0tr:
## Summary by Sourcery

Simplify and centralize controller image handling by extracting version and variant logic into utilities and enabling environment-driven image variant configuration

New Features:
- Allow overriding the controller image variant via the MODEL_RUNNER_CONTROLLER_VARIANT environment variable

Enhancements:
- Add controller_image.go with helpers for determining image version, variant, and formatted image tags
- Refactor EnsureControllerImage, PruneControllerImages, and CreateControllerContainer to use the new image naming utilities
- Remove inline tag constants and logic in favor of the centralized helper functions

### DIFF
--- a/cmd/cli/pkg/standalone/controller_image.go
+++ b/cmd/cli/pkg/standalone/controller_image.go
@@ -1,0 +1,48 @@
+package standalone
+
+import (
+	"os"
+
+	gpupkg "github.com/docker/model-runner/cmd/cli/pkg/gpu"
+)
+
+const (
+	// ControllerImage is the image used for the controller container.
+	ControllerImage = "docker/model-runner"
+	// defaultControllerImageVersion is the image version used for the controller container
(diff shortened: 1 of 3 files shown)


cmd/cli/pkg/standalone/controller_image.go
```diff
@@ -0,0 +1,48 @@
...
+	case gpupkg.GPUSupportCUDA:
+		return "cuda"
+	default:
+		return ""
+	}
+}
+
+func fmtControllerImageName(repo, version, variant string) string {
+	tag := repo + ":" + version
+	if len(variant) > 0 {
```
doringeman, line 40: nit:
```suggestion
	if variant != "" {
```
I guess the compiler produces the same output nowadays.

PR #216, opened 2025-10-11
Add start-runner, restart-runner and stop-runner commands

ericcurtin:
To start, restart and stop container without pulling or removing the image.


ericcurtin:
@doringeman start/stop is supposed to be like install or uninstall without:

docker pull

and

docker rmi -f

if that makes sense, this is incomplete though.

cmd/cli/commands/stop-runner.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"github.com/docker/model-runner/cmd/cli/commands/completion"
+	"github.com/spf13/cobra"
+)
+
+func newStopRunner() *cobra.Command {
+	var models bool
+	c := &cobra.Command{
+		Use:   "stop-runner",
+		Short: "Stop Docker Model Runner",
```
ilopezluna:
```suggestion
		Short: "Stop Docker Model Runner (Docker Engine only)",
```

.gitignore
```diff
@@ -2,6 +2,7 @@
 model-distribution-tool
 model-runner
 model-runner.sock
+docker-model
```
ilopezluna, line 5: What is `docker-model`?

cmd/cli/commands/stop-runner.go
ericcurtin: Sprinkling more of these around thanks

PR #218, opened 2025-10-13
Add vLLM support

doringeman:
## Summary by Sourcery

Add support for the vLLM backend in the Docker build and CI pipeline, reorganize the Dockerfile for selectable backends via the BACKEND argument, and centralize container setup with a shared setup-modelrunner script.

New Features:
- Add vLLM backend support with BACKEND and VLLM_VERSION build arguments and a dedicated final-vllm stage
- Introduce setup-modelrunner.sh script to unify modelrunner user creation and permission setup

Build:
- Restructure Dockerfile into multi-stage builds with separate final stages for llama.cpp and vLLM backends

CI:
- Extend GitHub Actions release workflow to build, tag, and push vLLM CUDA images

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,6 +19,11 @@ on:
         required: false
         type: string
         default: "latest"
+      buildCuda:
+        description: 'Build CUDA variant'
+        required: false
+        type: boolean
+        default: true
 
 jobs:
   test:
@@ -93,6 +98,7 @@ jobs:
           tags: ${{ steps.tags.outputs.cpu }}
 
       - name: Build CUDA image
+        if: ${{ inputs.buildCuda }}
         uses: docker/build-push-action@v5
         with:
           file: Dockerfile


.github/workflows/release.yml
```diff
@@ -99,3 +111,16 @@ jobs:
           sbom: true
           provenance: mode=max
           tags: ${{ steps.tags.outputs.cuda }}
+
+      - name: Build vLLM CUDA image
```
ilopezluna, line 115: Actually I have same question, why don't you add the conditional check here?
`${{ inputs.buildCuda }}`

.github/workflows/release.yml
```diff
@@ -99,3 +111,16 @@ jobs:
...
+        uses: docker/build-push-action@v5
+        with:
+          file: Dockerfile
+          platforms: linux/amd64, linux/arm64
+          build-args: |
+            "BACKEND=vllm"
+            "VLLM_VERSION=v0.11.0"
```
ilopezluna, line 122: `VLLM_VERSION` should we also be an input field?

Dockerfile
```diff
@@ -4,6 +4,8 @@ ARG GO_VERSION=1.24
 ARG LLAMA_SERVER_VERSION=latest
 ARG LLAMA_SERVER_VARIANT=cpu
 ARG LLAMA_BINARY_PATH=/com.docker.llama-server.native.linux.${LLAMA_SERVER_VARIANT}.${TARGETARCH}
+ARG BACKEND=llamacpp
+ARG VLLM_VERSION=v0.11.0
```
ilopezluna, line 8: Why not latest?

.github/workflows/release.yml
doringeman: Yes.

.github/workflows/release.yml
doringeman: I will totally remove this, I don't think we'll really need it.

Dockerfile
ericcurtin: This is the latest version :)

https://github.com/vllm-project/vllm/releases/tag/v0.11.0

I think @doringeman is right here, we want the latest released version, latest on the git branch is likely often broken. We are GA now, auto-bumping to latest is probably a bad idea, you need bulletproof CI for that and even then you should bump manually in a PR so we get CI runs.

doringeman: Pinned it for test for now, I will switch to `latest`, thanks!

doringeman:
Moved to https://github.com/docker/model-runner/pull/220 so I can run GH Actions.

PR #221, opened 2025-10-13
Add vLLM

doringeman:
[Building](https://github.com/docker/model-runner/actions/runs/18503681762) `v0.0.14-rc1`.
[Building](https://github.com/docker/model-runner/actions/runs/18710578381) `v0.0.14-rc2`.

Make sure you're using a Docker Engine context with `nvidia` runtime.

Build:
```
docker build -t docker/model-runner:test-vllm-cuda \
    --target final-vllm \
    --build-arg LLAMA_SERVER_VARIANT=cuda \
    --build-arg BASE_IMAGE=nvidia/cuda:12.9.0-runtime-ubuntu24.04 \
    --build-arg VLLM_VERSION=0.11.0 --platform linux/amd64 \
    .
```

Run:
```
docker run --rm -p 8080:8080 -e MODEL_RUNNER_PORT=8080 --gpus all --runtime=nvidia docker/model-runner:test-vllm-cuda
```

Test:
```
MODEL_RUNNER_HOST=http://localhost:8080 docker model run --backend vllm aistaging/smollm2-vllm hi
```

Loading the model is expected to take ~30s on Docker Offload.
E.g.,
```
$ MODEL_RUNNER_HOST=http://localhost:8080 docker model run --backend vllm aistaging/smollm2-vllm hi
Unable to find model 'aistaging/smollm2-vllm' locally. Pulling from the server.
Downloaded 272.05MB of 272.05MB
Model pulled successfully
Hello there! I’m here to help with an interesting task. What can I assist you with?
```

---
## Summary by Sourcery

Add support for building and publishing a vLLM-enabled CUDA Docker image by extending the Dockerfile and GitHub Actions release workflow.

New Features:
- Introduce a new vLLM variant in the Dockerfile that installs uv and the specified vLLM version
- Add vllmVersion input and corresponding tagging logic in the release workflow to handle vLLM CUDA images
- Add a build-and-push step for the vLLM CUDA image in GitHub Actions using docker/build-push-action

## Summary by Sourcery

Add vLLM backend support to model-runner, including backend implementation, scheduler integration, Docker build for vLLM CUDA images, CLI updates, and associated tests.

New Features:
- Introduce vLLM backend with installation, run, status, disk usage, and memory estimation support
- Automatically select vLLM backend for safetensors-format models in the scheduler
- Add a vLLM-CUDA Docker image variant in Dockerfile and GitHub Actions release workflow
- Extend CLI to recognize and document the vLLM backend as a valid option

Enhancements:
- Sanitize backend arguments for safe logging and consolidate format compatibility logic in distribution client
- Add platform.SupportsVLLM utility and refactor runner and loader to pass modelRef parameter
- Upgrade GitHub Actions references to fixed commit SHAs across workflows

Build:
- Add Dockerfile target final-vllm and supporting environment setup for vLLM

CI:
- Add build-and-push step for the vLLM CUDA image and update CI workflows with action version pins

Documentation:
- Update CLI reference documentation to include vLLM in backend options

Tests:
- Add unit tests for vLLM configuration argument generation
- Extend distribution client tests for platform-dependent safetensors support

### DIFF
--- a/pkg/inference/backends/vllm/vllm_config.go
+++ b/pkg/inference/backends/vllm/vllm_config.go
@@ -28,13 +28,13 @@ func (c *Config) GetArgs(bundle types.ModelBundle, socket string, mode inference
 	args := append([]string{}, c.Args...)
 
 	// Add the serve command and model path (use directory for safetensors)
-	modelPath := filepath.Dir(bundle.SafetensorsPath())
-	if modelPath != "" {
-		// vLLM expects the directory containing the safetensors files
-		args = append(args, "serve", modelPath)
-	} else {
+	safetensorsPath := bundle.SafetensorsPath()
+	if safetensorsPath == "" {
 		return nil, fmt.Errorf("safetensors path required by vLLM backend")
 	}
+	modelPath := filepath.Dir(safetensorsPath)
+	// vLLM expects the directory containing the safetensors files
+	args = append(args, "serve", modelPath)
 
 	// Add socket arguments
 	args = append(args, "--uds", socket)
--- a/pkg/inference/backends/vllm/vllm_config_test.go
+++ b/pkg/inference/backends/vllm/vllm_config_test.go
@@ -38,11 +38,21 @@ func (m *mockModelBundle) RootDir() string {
 
 func TestGetArgs(t *testing.T) {
 	tests := []struct {
-		name     string
-		config   *inference.BackendConfiguration
-		bundle   *mockModelBundle
-		expected []string
+		name        string
+		config      *inference.BackendConfiguration
+		bundle      *mockModelBundle
+		expected    []string
+		expectError bool
 	}{
+		{
+			name: "empty safetensors path should error",
+			bundle: &mockModelBundle{
+				safetensorsPath: "",
+			},
+			config:      nil,
+			expected:    nil,
+			expectError: true,
+		},
 		{
 			name: "basic args without context size",
 			bundle: &mockModelBundle{
@@ -116,6 +126,14 @@ func TestGetArgs(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			config := NewDefaultVLLMConfig()
 			args, err := config.GetArgs(tt.bundle, "/tmp/socket", inference.BackendModeCompletion, tt.config)
+
+			if tt.expectError {
+				if err == nil {
+					t.Fatalf("expected error but got none")
+				}
+				return
+			}
+
 			if err != nil {
 				t.Fatalf("unexpected error: %v", err)
 			}


.github/workflows/release.yml
```diff
@@ -19,6 +19,11 @@ on:
         required: false
         type: string
         default: "latest"
+      vllmVersion:
+        description: 'vLLM version'
+        required: false
+        type: string
+        default: "latest"
```
ericcurtin: 0.11.0 please, we are GA now.

Dockerfile
```diff
@@ -77,3 +77,21 @@ ENV LD_LIBRARY_PATH=/app/lib
...
 ENTRYPOINT ["/app/model-runner"]
+
+# --- vLLM variant ---
+FROM final AS vllm
+
+ARG VLLM_VERSION
+
+USER root
+
+RUN apt update && apt install -y python3 curl ca-certificates && rm -rf /var/lib/apt/lists/*
```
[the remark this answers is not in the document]
ericcurtin: --no-install-recommends isn't universally a good idea. I've seem important packages left out because of that

pkg/distribution/distribution/client.go
```diff
@@ -409,15 +409,5 @@ func checkCompat(image types.ModelArtifact) error {
 		return fmt.Errorf("config type %q is unsupported: %w", manifest.Config.MediaType, ErrUnsupportedMediaType)
 	}
 
-	// Check if the model format is supported
-	config, err := image.Config()
-	if err != nil {
-		return fmt.Errorf("reading model config: %w", err)
-	}
-
-	if config.Format == types.FormatSafetensors {
```
p1-0tr, line 426: nit: This will remain true on macOS and windows. So perhaps we should have a platform-specific check here? Also we could reverse the check to fail by default and succeed if `config.Format` is known (as a harmless future-proofing exercise).

pkg/inference/backends/vllm/vllm.go
```diff
@@ -36,33 +60,110 @@ func (v *vLLM) Name() string {
...
-// UsesExternalModelManagement implements
-// inference.Backend.UsesExternalModelManagement.
-func (l *vLLM) UsesExternalModelManagement() bool {
+func (v *vLLM) UsesExternalModelManagement() bool {
 	return false
 }
 
-// Install implements inference.Backend.Install.
 func (v *vLLM) Install(ctx context.Context, httpClient *http.Client) error {
-	// TODO: Implement.
```
p1-0tr, line 47: Perhaps we should verify `vllm` exists here, in case e.g. someone is running a DMR container with only `llama.cpp` inside it. Also platform checks would be good, since vllm will be Linux only.

pkg/inference/backends/vllm/vllm.go
```diff
@@ -36,33 +60,110 @@ func (v *vLLM) Name() string {
...
-	return errors.New("not implemented")
+	return nil
 }
 
-// Run implements inference.Backend.Run.
 func (v *vLLM) Run(ctx context.Context, socket, model string, mode inference.BackendMode, config *inference.BackendConfiguration) error {
-	// TODO: Implement.
-	v.log.Warn("vLLM backend is not yet supported")
```
p1-0tr, line 54: Still true on macOS and Windows.

pkg/distribution/distribution/client.go
doringeman: Done in https://github.com/docker/model-runner/pull/221/commits/42f472b7010c7847a5f811c089887b1adaed252f.

p1-0tr (approved):

Dockerfile
```diff
@@ -77,3 +74,31 @@ ENV LD_LIBRARY_PATH=/app/lib
...
+
+RUN apt update && apt install -y python3 python3-venv python3-dev curl ca-certificates build-essential && rm -rf /var/lib/apt/lists/*
+
+RUN mkdir -p /opt/vllm-env && chown -R modelrunner:modelrunner /opt/vllm-env
+
+USER modelrunner
+
+# Install uv and vLLM as modelrunner user
+RUN curl -LsSf https://astral.sh/uv/install.sh | sh \
+ && ~/.local/bin/uv venv --python /usr/bin/python3 /opt/vllm-env \
```
ericcurtin, line 93: If we change this to copy from the vllm/vllm-openai:v0.11.0 container we get DGX Spark support (I know I suggested doing it this less hacky way, apologies, didn't realize the container had aarch64 and this way doesn't appear to).

Could be a follow on PR too.

ericcurtin: Or this would be even better install the wheels from here:

https://wheels.vllm.ai/b8b302cde434df8c9289a2b465406b47ebab1c2d/vllm/

That commit sha is the 0.11.0 one.

They tipped me off in vLLM stack that they build CUDA x86_64 and aarch64 wheels for every commit. So this is the same thing, but has an aarch64 version also.

Be better than the hacky container copy (which is prone to error, missing files, OS mismatch, library version mismatch).

ericcurtin: This is a way to get that programatically:

```
$ git rev-list -n 1 v0.11.0
b8b302cde434df8c9289a2b465406b47ebab1c2d
```

pkg/inference/backends/vllm/vllm.go
```diff
@@ -36,33 +64,154 @@ func (v *vLLM) Name() string {
...
+	}
+
+	// Get arguments from config
+	args, err := v.config.GetArgs(bundle, socket, mode, backendConfig)
+	if err != nil {
+		return fmt.Errorf("failed to get vLLM arguments: %w", err)
+	}
+
+	// Add served model name
+	args = append(args, "--served-model-name", model, modelRef)
```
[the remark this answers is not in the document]
ilopezluna: it does not use coma separated values

ilopezluna:
@sourcery-ai review

ilopezluna:
@sourcery-ai review

pkg/inference/backends/vllm/vllm.go
```diff
@@ -3,16 +3,30 @@ package vllm
...
 	"github.com/docker/model-runner/pkg/logging"
+	"github.com/docker/model-runner/pkg/sandbox"
+	"github.com/docker/model-runner/pkg/tailbuffer"
 )
 
 const (
 	// Name is the backend name.
-	Name = "vllm"
+	Name    = "vllm"
+	vllmDir = "/opt/vllm-env/bin"
```
ericcurtin, line 29: I am sometimes curious why we care about the directory in the golang code. We could just add this directory to the PATH in the Dockerfile.

Could be useful if we start to use artifacts from this container for example for ROCm which expect things to be in a different path:

https://hub.docker.com/r/rocm/vllm

can worry about this in a follow on PR though

doringeman: I've done it like this to follow llama.cpp's approach where it can be installed on the host machine, where we can't have anything we want in the PATH.

pkg/inference/scheduling/scheduler.go
```diff
@@ -232,6 +214,42 @@ func (s *Scheduler) handleOpenAIInference(w http.ResponseWriter, r *http.Request
...
+			if config.Format == types.FormatSafetensors {
+				if vllmBackend, ok := s.backends[vllm.Name]; ok {
+					backend = vllmBackend
+				} else {
+					s.log.Warnf("Model %s is in safetensors format but vLLM backend is not available. "+
+						"Backend %s may not support this format and could fail at runtime.",
+						request.Model, backend.Name())
+				}
+			}
+		}
```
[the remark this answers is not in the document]
ilopezluna: the `--backend` will be removed in follow up pr

PR #230, opened 2025-10-15
New run prompt

ericcurtin:
Much more fully featured readline-like implementation which advanced keyboard usage.

## Summary by Sourcery

Provide an enhanced interactive "run" prompt by introducing a custom readline-like implementation with advanced keyboard shortcuts, multi-line input markers, history and command handling, while falling back to a basic input mode when not running in a terminal

New Features:
- Add generateInteractiveWithReadline for an advanced interactive mode with custom prompts, multi-line paste support, history and keyboard shortcuts
- Add generateInteractiveBasic as a fallback interactive mode for non-terminal environments
- Implement cmd/cli/readline package to manage terminal raw mode, input buffering, history persistence and cross-platform terminal control

Enhancements:
- Replace inline interactive loop in run command with terminal detection and modular interactive handlers
- Introduce slash commands (/help, /?, /bye, /exit) and unknown command feedback in interactive sessions
- Remove outdated static startup message from CLI docs to reflect new prompt behavior

Build:
- Update go.mod to adjust containerd/containerd and golang.org/x/crypto versions and add dependencies for readline functionality

Documentation:
- Remove old example lines about "Interactive chat mode started" in README and reference docs

### DIFF
--- a/cmd/cli/commands/run.go
+++ b/cmd/cli/commands/run.go
@@ -11,6 +11,7 @@ import (
 	"github.com/charmbracelet/glamour"
 	"github.com/docker/model-runner/cmd/cli/commands/completion"
 	"github.com/docker/model-runner/cmd/cli/desktop"
+	"github.com/docker/model-runner/cmd/cli/readline"
 	"github.com/fatih/color"
 	"github.com/spf13/cobra"
 	"golang.org/x/term"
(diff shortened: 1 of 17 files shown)


p1-0tr:
Seems we may be heading toward a good moment to revive - https://github.com/docker/model-cli/pull/144 ?

ericcurtin:
Genuinely I'm happy to continue here, we don't need 65k lines of code for a simple feature like this. Have you tested both to see what the UX is like?

cmd/cli/commands/run.go
```diff
@@ -81,6 +82,167 @@ func readMultilineInput(cmd *cobra.Command, scanner *bufio.Scanner) (string, err
...
+		fmt.Fprintln(os.Stderr, "  Ctrl + a            Move to the beginning of the line (Home)")
+		fmt.Fprintln(os.Stderr, "  Ctrl + e            Move to the end of the line (End)")
+		fmt.Fprintln(os.Stderr, "   Alt + b            Move back (left) one word")
+		fmt.Fprintln(os.Stderr, "   Alt + f            Move forward (right) one word")
+		fmt.Fprintln(os.Stderr, "  Ctrl + k            Delete the sentence after the cursor")
+		fmt.Fprintln(os.Stderr, "  Ctrl + u            Delete the sentence before the cursor")
+		fmt.Fprintln(os.Stderr, "  Ctrl + w            Delete the word before the cursor")
+		fmt.Fprintln(os.Stderr, "")
+		fmt.Fprintln(os.Stderr, "  Ctrl + l            Clear the screen")
+		fmt.Fprintln(os.Stderr, "  Ctrl + c            Stop the model from responding")
```
doringeman, line 108: I guess this shouldn't exit the interactive chat, right? Because it exists it in my iTerm2.

cmd/cli/commands/run.go
```diff
@@ -81,6 +82,167 @@ func readMultilineInput(cmd *cobra.Command, scanner *bufio.Scanner) (string, err
...
+		fmt.Fprintln(os.Stderr, `Use """ to begin a multi-line message.`)
+		fmt.Fprintln(os.Stderr, "")
+	}
+
+	usageShortcuts := func() {
+		fmt.Fprintln(os.Stderr, "Available keyboard shortcuts:")
+		fmt.Fprintln(os.Stderr, "  Ctrl + a            Move to the beginning of the line (Home)")
+		fmt.Fprintln(os.Stderr, "  Ctrl + e            Move to the end of the line (End)")
+		fmt.Fprintln(os.Stderr, "   Alt + b            Move back (left) one word")
+		fmt.Fprintln(os.Stderr, "   Alt + f            Move forward (right) one word")
```
doringeman, line 102: alt (option) + b/f don't on my iTerm2 on macOS.
option + arrow left/right does.

cmd/cli/commands/run.go
ericcurtin: Yes in this and current implementation we exist the interactive chat in Ctlr-C.

I agree we should try and change that, I'd like to do that as a follow on feature to this PR.

p1-0tr:
> Genuinely I'm happy to continue here, we don't need 65k lines of code for a simple feature like this. Have you tested both to see what the UX is like?

IIRC most of that change was `vendor` code.

Edit: yep, on our side there is ~1k lines of code in that PR

cmd/cli/commands/run.go
doringeman: * if I press it while the response is streaming.

cmd/cli/commands/run.go
ericcurtin: Thanks for testing

Did it work before?

Is is just specific to iTerm2?

ericcurtin:
> > Genuinely I'm happy to continue here, we don't need 65k lines of code for a simple feature like this. Have you tested both to see what the UX is like?
> 
> IIRC most of that change was `vendor` code.

Did you test this implementation vs that one? The difference is drastic.

cmd/cli/commands/run.go
ericcurtin: Opening an issue

cmd/cli/commands/run.go
```diff
@@ -81,6 +82,167 @@ func readMultilineInput(cmd *cobra.Command, scanner *bufio.Scanner) (string, err
...
+		fmt.Fprintln(os.Stderr, "  Ctrl + k            Delete the sentence after the cursor")
+		fmt.Fprintln(os.Stderr, "  Ctrl + u            Delete the sentence before the cursor")
+		fmt.Fprintln(os.Stderr, "  Ctrl + w            Delete the word before the cursor")
+		fmt.Fprintln(os.Stderr, "")
+		fmt.Fprintln(os.Stderr, "  Ctrl + l            Clear the screen")
+		fmt.Fprintln(os.Stderr, "  Ctrl + c            Stop the model from responding")
+		fmt.Fprintln(os.Stderr, "  Ctrl + d            Exit (/bye)")
```
doringeman, line 109:
```suggestion
		fmt.Fprintln(os.Stderr, "  Ctrl + d            Delete the character under the cursor. If the input line is empty, exit the chat (/bye)")
```

cmd/cli/commands/run.go
ericcurtin: https://github.com/docker/model-runner/issues/233

p1-0tr:
> Did you test this implementation vs that one? The difference is drastic.

Not yet, I wasn't assuming they do the same thing. More that if we are aiming for a richer CLI interface we could benefit from using sth like bubbletea.

ericcurtin:
> > Did you test this implementation vs that one? The difference is drastic.
> 
> Not yet, I wasn't assuming they do the same thing. More that if we are aiming for a richer CLI interface we could benefit from using sth like bubbletea.

The bubbletea PR is badly broken, tried bubbletea and huh... Not impressed

ericcurtin:
> > Genuinely I'm happy to continue here, we don't need 65k lines of code for a simple feature like this. Have you tested both to see what the UX is like?
> 
> IIRC most of that change was `vendor` code.
> 
> Edit: yep, on our side there is ~1k lines of code in that PR

It's still important to stay lean binary size and dependancies are not 100% free, still require maintenance, large dependancies increase the chance of CVE's appearing.

cmd/cli/commands/run.go
ericcurtin: https://github.com/docker/model-runner/issues/233

p1-0tr:
TBH I don't think binary size should differ that much (plus it's golang, so the lean binary ship has long sailed 🙈). But if we don't believe bubble tea fits our bill, that's fair enough for me.

cmd/cli/commands/run.go
ericcurtin: https://github.com/docker/model-runner/issues/234

doringeman: I searched for it and It's specific to macOS, you need an additional step to make it work - to enable `option` to be Meta/+Esc. So it's fine, don't bother in this PR.

cmd/cli/commands/run.go
ilopezluna: I just tried and it's actually the same as `/bye`

ericcurtin: This makes sense! Ctrl-d is exit prompt in all shells, I guess you can't type /bye mid-prompt which this is highlighting.

ericcurtin:
I think we should merge though because of better UX.

I do appreciate the issues we are finding though, I think we should fill up our issue tracker with these.

doringeman (approved):

PR #231, opened 2025-10-15
Add --detach flag to docker model run command

ericcurtin:
So we can load models in the background without interaction

## Summary by Sourcery

Add a `--detach`/`-d` option to `docker model run` to load models in the background without interaction, update command behavior, and include supporting tests and documentation.

New Features:
- Introduce `--detach`/`-d` flag on the run command to enable background model loading without user input.

Enhancements:
- Adjust run command logic to skip stdin reading and perform a silent preload when the `--detach` flag is set.

Documentation:
- Document the `--detach` flag in the reference and provide an example for pre-loading models.

Tests:
- Add `TestRunCmdDetachFlag` to validate the presence, default state, type, and behavior of the `--detach` flag.

### DIFF
--- a/cmd/cli/commands/run.go
+++ b/cmd/cli/commands/run.go
@@ -307,6 +307,7 @@ func newRunCmd() *cobra.Command {
 	var backend string
 	var ignoreRuntimeMemoryCheck bool
 	var colorMode string
+	var detach bool
 
 	const cmdArgs = "MODEL [PROMPT]"
 	c := &cobra.Command{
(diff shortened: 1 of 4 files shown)


p1-0tr (approved):

cmd/cli/commands/run.go
```diff
@@ -381,6 +385,21 @@ func newRunCmd() *cobra.Command {
 				}
 			}
 
+			// Handle --detach flag: just load the model without interaction
+			if detach {
+				// Make a minimal request to load the model into memory
+				err := desktopClient.Chat(backend, model, "", apiKey, func(content string) {
```
ilopezluna, line 391: The request will be recorded, not a big deal, but to keep in mind

ericcurtin: Makes me think do we have a way of forking without chatting? 🤔

cmd/cli/commands/run.go
```diff
@@ -381,6 +385,21 @@ func newRunCmd() *cobra.Command {
 				}
 			}
 
+			// Handle --detach flag: just load the model without interaction
+			if detach {
```
doringeman, line 389: Couldn't this have been placed at the beginning of the function so we don't need another `if !detach`?

PR #236, opened 2025-10-15
runner: Make runtime memory checking opt-in

p1-0tr:
The current model memory requirement checking routines are limited to llama.cpp, and because a separate library is used to do the estimation it is likely to lag behind llama.cpp in terms of supported models. This significantly complicates delivering a good experience for users of the feature. So make it opt-in, for now.

## Summary by Sourcery

Make runtime model memory requirement checking opt-in via an environment variable

New Features:
- Introduce MODEL_RUNNER_RUNTIME_MEMORY_CHECK environment variable to enable runtime memory checks

Enhancements:
- Add memory/settings package with SetRuntimeMemoryCheck and RuntimeMemoryEnabled to manage the check flag
- Guard the model memory estimation logic behind the runtime memory check toggle

### DIFF
--- a/main.go
+++ b/main.go
@@ -103,6 +103,10 @@ func main() {
 		log.Fatalf("unable to initialize %s backend: %v", llamacpp.Name, err)
 	}
 
+	if os.Getenv("MODEL_RUNNER_RUNTIME_MEMORY_CHECK") == "1" {
+		memory.SetRuntimeMemoryCheck(true)
+	}
+
 	memEstimator.SetDefaultBackend(llamaCppBackend)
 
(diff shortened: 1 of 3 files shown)


ericcurtin:
Is this opt-in everywhere? I think this is a good change, memory checking is hard to get right without hurting people in unprecedented situations. There's a crazy amount of factors where memory checking can be inaccurate, regardless of the the implementation.

p1-0tr:
I'll have to follow up with a PR for DD to allow people to turn it back on, with this patch it will default to being turned off.

PR #240, opened 2025-10-15
Add model name normalization and display stripping

ericcurtin:
Add normalization to configure, list, and unload commands

## Summary by Sourcery

Introduce consistent model name normalization and display stripping across the CLI and manager, ensuring default organization and tag are applied or removed appropriately.

New Features:
- Add normalizeModelName to prepend "ai/" and append ":latest" to model names when missing.
- Add stripDefaultsFromModelName to remove default "ai/" prefix and ":latest" tag for user-facing display.

Enhancements:
- Integrate normalization into configure, list, pull, push, inspect, remove, unload, run, tag, and ps CLI commands.
- Replace normalizeHuggingFaceModelName with normalizeModelName in the desktop client and HTTP handlers.
- Normalize model names in the manager’s create, get, delete, and OpenAI endpoints.
- Strip default prefixes and tags in CLI list and ps outputs for cleaner display.

Tests:
- Add unit tests covering normalizeModelName and stripDefaultsFromModelName for various naming scenarios.

### DIFF
--- a/cmd/cli/commands/utils.go
+++ b/cmd/cli/commands/utils.go
@@ -11,6 +11,11 @@ import (
 )
 
 const (
+	defaultOrg = "ai"
+	defaultTag = "latest"
+)
+
+const (
 	enableViaCLI = "Enable Docker Model Runner via the CLI → docker desktop enable model-runner"
 	enableViaGUI = "Enable Docker Model Runner via the GUI → Go to Settings->AI->Enable Docker Model Runner"
 )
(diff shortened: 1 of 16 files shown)


ericcurtin:
Backend and frontend changes to make "ai/" and ":latest" optional and defaults, we need both. Now it's just:

docker model run gemma

cmd/cli/desktop/desktop.go
```diff
@@ -66,6 +70,58 @@ func normalizeHuggingFaceModelName(model string) string {
...
 }
 
+// normalizeModelName adds the default organization prefix (ai/) and tag (:latest) if missing.
+// Examples:
+//   - "gemma3" -> "ai/gemma3:latest"
+//   - "gemma3:v1" -> "ai/gemma3:v1"
+//   - "myorg/gemma3" -> "myorg/gemma3:latest"
+//   - "ai/gemma3:latest" -> "ai/gemma3:latest" (unchanged)
+//   - "hf.co/model" -> "hf.co/model" (unchanged - has registry)
+func normalizeModelName(model string) string {
```
p1-0tr: Could we have one `normalizeModelName`?

cmd/cli/commands/utils.go
```diff
@@ -32,3 +37,25 @@ func handleNotRunningError(err error) error {
...
 }
+
+// stripDefaultsFromModelName removes the default "ai/" prefix and ":latest" tag for display.
+// Examples:
+//   - "ai/gemma3:latest" -> "gemma3"
+//   - "ai/gemma3:v1" -> "ai/gemma3:v1"
+//   - "myorg/gemma3:latest" -> "myorg/gemma3"
+//   - "gemma3:latest" -> "gemma3"
+//   - "hf.co/bartowski/model:latest" -> "hf.co/bartowski/model"
+func stripDefaultsFromModelName(model string) string {
```
p1-0tr, line 48: nit: Maybe this should be implemented in the model manager along with `NormalizeModelName`? Could be called `ExtractModelBaseName()`.

p1-0tr (approved):
One small nit, but otherwise LGTM.

cmd/cli/commands/utils.go
ericcurtin: Good consideration, I thought of that at one point, might do it in a follow on PR. The reason it's not is one has wider use than the other. But I get your point they are similar, one does the opposite of the other.

I actually think it's a pity we didn't go lowercase for our dmr quantization sometimes, the :tag . We actually have code here that converts everything to lowercase for huggingface and it's generally what ollama does also.

Issue #259, opened 2025-10-19
High CPU usage of docker-model.exe without any models running


ThePlenkov:
Hi!! Could you please explain why this happens:

<img width="796" height="601" alt="Image" src="https://github.com/user-attachments/assets/bb6bc860-d0c0-4121-a4c8-b971bc8d7e6b" />

My PC is using max of CPU

but when I go to docker model ps - there is nothing there

Coould you please advice why and how to investigate the reason?

thanks!

ThePlenkov:
```

C:\Users\pplenkov>docker model unload --all
No models are running.

C:\Users\pplenkov>docker model status
Docker Model Runner is running

Status:
llama.cpp: running llama.cpp latest-cuda (sha256:ba7f4e5adf0a03e222cc3f73c1cf811e348c9de547b431359d9a417ec0babe76) version: ca71fb9

C:\Users\pplenkov>docker model ps
MODEL NAME  BACKEND  MODE  LAST USED

C:\Users\pplenkov>
```

f2bo:
I have the same problem and can consistently reproduce it (in Docker Desktop for Windows). 

1. Open Task Manager to monitor running processes
1. Launch Docker Desktop and open its Dashboard
1. Switch to the **Models** section
1. Select the **Logs** tab and notice that a **docker-model.exe** process is launched
1. Switch back to the **Local** tab
1. Repeat the process several times switching back and forth between the **Logs** tab and a different tab. Notice that a new instance of **docker-model.exe** is created each time the **Logs** tab is selected and that each one consumes approximately 6% of the CPU.

In my case, the last entry shown by the logs is from 2 months ago even though I have used Model Runner since then.

The same happens whenever I query the logs from the command line using `docker model logs`. However, in this case, I notice that the command hangs and I need to press CTRL + C to regain control. Note that I did not specify the `--follow` argument. If instead I query the logs using `docker model logs --no-engines`, the last entry displayed is from today and the command exits successfully.

ericcurtin:
@p1-0tr do you have cycles to look at this?

f2bo:
I investigated the issue I mentioned previously regarding the logs a little bit more. After determining that, in Windows, log files are stored at `C:\Users\<User>\AppData\Local\Docker\log\host`, I noticed that it contained several related files:

- `inference-llama.cpp-server.log`
- `inference-llama.cpp-server.log.0`
- `inference-llama.cpp-server.log.2`

[...]

PR #260, opened 2025-10-19
Fix CPU spin when defunct runner is still in use

ericcurtin:
User reported this problem on windows.

## Summary by Sourcery

Bug Fixes:
- Adjust idleCheckDuration to only trigger immediate checks for unused defunct runners, avoiding tight loops when runners are still active


ericcurtin:
@xenoscopic @doringeman do you think this will fix https://github.com/docker/model-runner/issues/259 ?

doringeman:
Have you been able to reproduce the issue @ericcurtin?

doringeman:
I don't think it will fix it. It'll postpone the eviction of a runner that's unusable if there are more references to it.

ericcurtin:
I didn't reproduce, closing...

Issue #139, opened 2025-08-24
Pulling models from behind firewall not possible even though pulling images works


holger-stenzhorn:
I am running Docker Engine on a Rocky Linux VM behind a firewall and therefore I have provided a proxy configuration in `/etc/systemd/system/docker.service.d/http-proxy.conf`:

```
[Service]
Environment="HTTP_PROXY=http://some.proxy.server:3128"
Environment="HTTPS_PROXY=http://some.proxy.server:3128"
Environment="NO_PROXY=localhost,127.0.0.1"
```

With this configuration pulling of images works BUT pulling of models does not:

```
❯ docker model run ai/smollm2
353d9a3ec7ee: Pull complete Digest: sha256:00fba948428db7f9a750a7a74b3af55ab99801d0a6f1a1d7b68169a948661342
Status: Downloaded newer image for docker/model-runner:latest

Successfully pulled docker/model-runner:latest
Starting model runner container docker-model-runner...
Unable to find model 'ai/smollm2' locally. Pulling from the server.
Failed to pull model: pulling ai/smollm2 failed with status 500 Internal Server Error: error while pulling model: reading model from registry: failed to pull model "ai/smollm2": UNKNOWN - Get "https://index.docker.io/v2/": net/http: TLS handshake timeout
```

So seemingly the proxy configuration is not passed to the container `docker-model-runner` but there is also no possibility to manually specify it.

xenoscopic:
Thanks for filing the report!

@thaJeztah do you know if there's an easy way for us to pass the Docker Engine HTTP proxy setting into our model runner container in Docker CE?  I guess we could bind-mount / parse `http-proxy.conf`, but I don't know if the permissions / ownership on that file would allow it.  Our other option would probably be adding a `--proxy` flag to `docker model install-runner`, but it would be nice from an integration standpoint if it could be inherited from `dockerd`'s config.

ericcurtin:
@holger-stenzhorn just curious we are putting in effort to make contributing to docker model-runner more contributor friendly, would you be interested in working on this?

holger-stenzhorn:
@ericcurtin Yes, that sounds quite interesting for me.

ericcurtin:
> [@ericcurtin](https://github.com/ericcurtin) Yes, that sounds quite interesting for me.

Might be worth going through docker code also, ideally we be as most similar to docker as possible, (just use their config stuff)

ericcurtin:
[...]

PR #263, opened 2025-10-19
Add proxy support for model pulling behind firewall

ericcurtin:
- Pass proxy environment variables (HTTP_PROXY, HTTPS_PROXY, NO_PROXY) to docker-model-runner container
- Configure HTTP transport to use ProxyFromEnvironment for model pulling
- Add tests for proxy configuration in both container creation and transport

## Summary by Sourcery

Enable HTTP(S) proxy support for model pulling behind a firewall by configuring the HTTP transport with ProxyFromEnvironment and passing proxy environment variables into spawned containers, with accompanying unit tests.

New Features:
- Enable HTTP(S) proxy support for model pulling via ProxyFromEnvironment in the HTTP transport

Enhancements:
- Pass proxy environment variables (HTTP_PROXY, HTTPS_PROXY, NO_PROXY and lowercase equivalents) into the docker-model-runner container

Tests:
- Add unit tests for proxy-aware HTTP transport configuration
- Add unit tests to verify proxy environment variables are correctly passed to container creation

### DIFF
--- a/cmd/cli/pkg/standalone/containers.go
+++ b/cmd/cli/pkg/standalone/containers.go
@@ -230,6 +230,14 @@ func CreateControllerContainer(ctx context.Context, dockerClient *client.Client,
 	if doNotTrack {
 		env = append(env, "DO_NOT_TRACK=1")
 	}
+	
+	// Pass proxy environment variables to the container if they are set
+	proxyEnvVars := []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY", "http_proxy", "https_proxy", "no_proxy"}
+	for _, proxyVar := range proxyEnvVars {
+		if value, ok := os.LookupEnv(proxyVar); ok {
+			env = append(env, proxyVar+"="+value)
+		}
+	}
 	config := &container.Config{
 		Image: imageName,
 		Env:   env,
--- a/main.go
+++ b/main.go
@@ -70,12 +70,22 @@ func main() {
 
 	memEstimator := memory.NewEstimator(sysMemInfo)
 
+	// Create a proxy-aware HTTP transport
+	// Use a safe type assertion with fallback, and explicitly set Proxy to http.ProxyFromEnvironment
+	var baseTransport *http.Transport
+	if t, ok := http.DefaultTransport.(*http.Transport); ok {
+		baseTransport = t.Clone()
+	} else {
+		baseTransport = &http.Transport{}
+	}
+	baseTransport.Proxy = http.ProxyFromEnvironment
+
 	modelManager := models.NewManager(
 		log,
 		models.ClientConfig{
 			StoreRootPath: modelPath,
 			Logger:        log.WithFields(logrus.Fields{"component": "model-manager"}),
-			Transport:     resumable.New(http.DefaultTransport),
+			Transport:     resumable.New(baseTransport),
 		},
 		nil,
 		memEstimator,


ericcurtin:
@holger-stenzhorn, @xenoscopic should this close:

https://github.com/docker/model-runner/issues/139

?

holger-stenzhorn:
> @holger-stenzhorn, @xenoscopic should this close:
> 
> 
> 
> https://github.com/docker/model-runner/issues/139
> 
> 
> 
> ?
> 
> 

@ericcurtin: Yes, that should do the trick. I will test it as soon as I can today. Thanks a lot for your support.

cmd/cli/pkg/standalone/containers_test.go
```diff
@@ -0,0 +1,122 @@
+package standalone
+
+import (
+	"os"
+	"testing"
+)
+
+// TestProxyEnvironmentVariablesPassedToContainer verifies that proxy environment
+// variables are correctly identified and would be passed to the container.
+func TestProxyEnvironmentVariablesPassedToContainer(t *testing.T) {
```
doringeman: I don't understand the need of this test. This doesn't test any production code, does it?
If I remove the logic you're adding in this PR in `cmd/cli/pkg/standalone/containers.go`, the test will still pass.
```
	// Pass proxy environment variables to the container if they are set
	proxyEnvVars := []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY", "http_proxy", "https_proxy", "no_proxy"}
	for _, proxyVar := range proxyEnvVars {
		if value := os.Getenv(proxyVar); value != "" {
			env = append(env, proxyVar+"="+value)
		}
	}
```

doringeman (changes requested):

cmd/cli/pkg/standalone/containers.go
```diff
@@ -230,6 +230,14 @@ func CreateControllerContainer(ctx context.Context, dockerClient *client.Client,
 	if doNotTrack {
 		env = append(env, "DO_NOT_TRACK=1")
 	}
+	
+	// Pass proxy environment variables to the container if they are set
+	proxyEnvVars := []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY", "http_proxy", "https_proxy", "no_proxy"}
+	for _, proxyVar := range proxyEnvVars {
+		if value := os.Getenv(proxyVar); value != "" {
```
p1-0tr: nit: `LookupEnv` may be a bit more accurate. Though I assume none of those variables make sense with an empty value.

ericcurtin: @p1-0tr could you further explain? It's unclear to be what advantage this would bring, or maybe something else is being suggested?

```
-     if value := os.Getenv(proxyVar); value != "" {
+     if value, ok := os.LookupEnv(proxyVar); ok && value != "" {
```

ericcurtin:
@doringeman test removed, please re-review

cmd/cli/pkg/standalone/containers.go
p1-0tr: I meant that:

    -     if value := os.Getenv(proxyVar); value != "" {
    +     if value, ok := os.LookupEnv(proxyVar); ok {

would be "more correct" in terms of how env variables work in general. It probably does not change much in the context of the variables we are forwarding (they likely are never used with an empty value), but I can't say for sure OTOH (so I would lean toward the more generic logic by default).

ericcurtin: The way we did it is the curl_getenv way.

We need this PR unblocked, lets change

p1-0tr (approved):

PR #266, opened 2025-10-20
Add /api/generate endpoint for model loading and unloading

ericcurtin:
So we can load and unload models

### DIFF
--- a/cmd/cli/commands/run.go
+++ b/cmd/cli/commands/run.go
@@ -691,8 +691,12 @@ func newRunCmd() *cobra.Command {
 				}
 			}
 
+			// Check if a prompt was explicitly provided (even if empty string)
+			// If args length > 1, then a prompt argument was provided (even if it's "")
+			explicitPromptProvided := len(args) > 1
+
 			// Handle --detach flag: just load the model without interaction
-			if detach {
+			if detach || (explicitPromptProvided && prompt == "") {
 				// Make a minimal request to load the model into memory
 				err := desktopClient.Chat(model, "", nil, func(content string) {
 					// Silently discard output in detach mode
@@ -713,6 +717,14 @@ func newRunCmd() *cobra.Command {
 				cmd.Println()
 				return nil
 			}
+
+			// Preload the model in the background to optimize for the first user interaction
+			// This makes sure the model is loaded when the user types their first prompt
+			go func() {
+				_ = desktopClient.Chat(model, "", nil, func(content string) {
+					// Silently preload the model - discard output
+				}, false)
+			}()
 
 			// Use enhanced readline-based interactive mode when terminal is available
 			if term.IsTerminal(int(os.Stdin.Fd())) {
--- a/cmd/cli/commands/stop.go
+++ b/cmd/cli/commands/stop.go
@@ -1,0 +1,48 @@
+package commands
+
+import (
+	"fmt"
+
+	"github.com/docker/model-runner/cmd/cli/commands/completion"
+	"github.com/docker/model-runner/cmd/cli/desktop"
+	"github.com/docker/model-runner/pkg/inference/models"
+	"github.com/spf13/cobra"
+)
+
+func newStopCmd() *cobra.Command {
+	var backend string
+
+	const cmdArgs = "MODEL"
+	c := &cobra.Command{
+		Use:   "stop " + cmdArgs,
+		Short: "Stop a running model",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			model := models.NormalizeModelName(args[0])
+			unloadResp, err := desktopClient.Unload(desktop.UnloadRequest{Backend: backend, Models: []string{model}})
+			if err != nil {
+				err = handleClientError(err, "Failed to stop model")
+				return handleNotRunningError(err)
+			}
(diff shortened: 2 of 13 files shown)


ericcurtin:
@doringeman @ilopezluna this has a vLLM optimization also, load the model as soon as you execute "docker model run" in the background while you are typing the prompt

cmd/cli/commands/stop.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/docker/model-runner/pkg/inference/models"
+	"github.com/spf13/cobra"
+)
+
+func newStopCmd() *cobra.Command {
+	var backend string
+
+	const cmdArgs = "MODEL"
+	c := &cobra.Command{
+		Use:   "stop " + cmdArgs,
```
doringeman, line 17: Why do we need this and don't use `unload`?

cmd/cli/commands/utils.go
```diff
@@ -39,6 +39,14 @@ func handleClientError(err error, message string) error {
...
 }
 
+// handleNotRunningError checks if the error indicates that the model was not running
+// and returns a user-friendly message in that case
+func handleNotRunningError(err error) error {
+	// For now, just return the error as-is
+	// This function can be expanded to handle specific "model not running" errors in the future
+	return err
+}
+
```
doringeman, line 49: This is done in the `handleClientError`.
```suggestion

```

pkg/inference/scheduling/scheduler.go
```diff
@@ -535,6 +535,122 @@ func (s *Scheduler) handleModels(w http.ResponseWriter, r *http.Request) {
 	s.modelManager.ServeHTTP(w, r)
 }
 
+// HandleGenerate handles /api/generate requests
+// If prompt is empty, loads the model into memory
+// If prompt is empty and keep_alive is 0, unloads the model
+func (s *Scheduler) HandleGenerate(w http.ResponseWriter, r *http.Request) {
```
doringeman, line 541: Why isn't this under `inference.InferencePrefix` like all the other scheduler endpoints?

pkg/internal/utils/log.go
```diff
@@ -49,3 +50,24 @@ func SanitizeForLog(s string) string {
...
+		// If the model name contains unsafe characters, clean it
+		// Replace unsafe characters with underscores
+		safeModelName := regexp.MustCompile(`[^a-zA-Z0-9._/-]`).ReplaceAllString(modelName, "_")
+		// Ensure it doesn't start or end with unsafe sequences like ".." that could be path traversal
+		safeModelName = strings.ReplaceAll(safeModelName, "..", "_")
+		return safeModelName
+	}
+	
+	return modelName
+}
```
doringeman, line 73: Dead code.
```suggestion

```

pkg/inference/scheduling/scheduler.go
```diff
@@ -535,6 +535,122 @@ func (s *Scheduler) handleModels(w http.ResponseWriter, r *http.Request) {
...
+				DoneReason: "unload",
+			}
+			
+			w.Header().Set("Content-Type", "application/json")
+			json.NewEncoder(w).Encode(response)
+			return
+		} else {
+			// Load the model by requesting a minimal inference
+			// This will trigger the loading mechanism in the loader
+			backend := s.defaultBackend
```
doringeman, line 589: Why does this work for the default backend only?

cmd/cli/commands/run.go
```diff
@@ -714,6 +718,14 @@ func newRunCmd() *cobra.Command {
 				return nil
 			}
 
+			// Preload the model in the background to optimize for the first user interaction
+			// This makes sure the model is loaded when the user types their first prompt
+			go func() {
+				_ = desktopClient.Chat(model, "", nil, func(content string) {
+					// Silently preload the model - discard output
```
doringeman, line 725: The request/response will still be recorder.
Why don't we just load the model without running inference?

cmd/cli/commands/stop.go
ilopezluna: I've found myself using `docker model stop` in the past, I think its exactly the same behavior than `docker model unload`, to me `docker model stop` its more intuitive.
If the goal of the `stop` command is to `unload` a model, maybe we could use an alias? so `docker model stop` and `docker model unload` do the same thing?

pkg/inference/scheduling/scheduler.go
ilopezluna, line 541: If I understand correctly, requests to `/api/generate` are not going to be tracked, right?
We should track same way as `/.../chat/completions`

cmd/cli/commands/stop.go
doringeman: Yes, an alias would be the way to go for this IMO.

pkg/inference/scheduling/scheduler.go
doringeman: I see it calls `s.handleOpenAIInference`, so it will be tracked. Am I missing something?

cmd/cli/commands/stop.go
ericcurtin: Agree, lets alias, might as well create "docker model load" also

PR #294, opened 2025-10-22
docker ps sometimes lists years ago

ericcurtin:
This should be more accurate

## Summary by Sourcery

Introduce an InUse indicator and leverage it in the ps command to accurately display backend usage status and avoid incorrect time computations.

New Features:
- Add an InUse flag to BackendStatus in both the CLI and scheduling API

Bug Fixes:
- Prevent ‘docker ps’ from showing large ‘years ago’ durations for backends with a zero LastUsed timestamp

Enhancements:
- Update the ps command to display ‘in use’ for active backends or as a fallback, and only compute humanized durations for idle backends


ericcurtin:
Have not

PR #299, opened 2025-10-23
Add MUSA Support

yeahdongcn:
### Testing Done

```bash
# Local build musa container image
❯ pwd
/home/xiaodongye/ws/ggml/model-runner
❯ docker build -t docker/model-runner:latest-musa -f llamacpp/musa.Dockerfile . 
❯ docker images | grep model-runner
docker/model-runner                                                       latest-musa                                b547f47e4c4a   15 hours ago    210MB

❯ pwd
/home/xiaodongye/ws/ggml/model-runner/cmd/cli
❯ make build
Building model-cli...
go build -ldflags="-s -w" -o model-cli .
❯ ./model-cli install-runner
Error: unable to initialize standalone model runner container: failed to create container docker-model-runner: Error response from daemon: No command specified
Usage:
  model install-runner [flags]

Flags:
      --do-not-track   Do not track models usage in Docker Model Runner
      --gpu string     Specify GPU support (none|auto|cuda|rocm|musa) (default "auto")
  -h, --help           help for install-runner
      --host string    Host address to bind Docker Model Runner (default "127.0.0.1")
      --port uint16    Docker container port for Docker Model Runner (default: 12434 for Docker Engine, 12435 for Cloud mode)

unable to initialize standalone model runner container: failed to create container docker-model-runner: Error response from daemon: No command specified
```

I guess the GPU auto-detection is working as expected, and the CLI is trying to fetch the `docker/model-runner:latest-musa image` (which should be built using the Dockerfiles from https://github.com/docker/model-runner/pull/298).

PR #302, opened 2025-10-23
Add all llamacpp dockerfiles

ericcurtin:
For the community

### DIFF
--- a/llamacpp/native/cann.Dockerfile
+++ b/llamacpp/native/cann.Dockerfile
@@ -1,0 +1,49 @@
+# syntax=docker/dockerfile:1
+
+ARG CANN_VERSION=8.0.0-910b
+ARG CANN_IMAGE_VARIANT=openeuler22.03
+
+FROM quay.io/ascend/cann:{CANN_VERSION}-{CANN_IMAGE_VARIANT}-py3.10 AS builder
+
+ARG TARGETARCH
+ARG CANN_IMAGE_VARIANT
+
+RUN apt-get update && apt-get install -y cmake ninja-build git build-essential curl
+
+WORKDIR /llama-server
+
+COPY .git .git
+COPY native/CMakeLists.txt .
+COPY native/src src
+COPY native/vendor vendor
+
+# Fix submodule .git file to point to correct location in container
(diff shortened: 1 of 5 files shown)


yeahdongcn:
I did the following updates to make the `docker/model-runner:latest-musa` correctly built:

```bash
❯ pwd
/home/xiaodongye/ws/ggml/model-runner
❯ docker build -t docker/model-runner:latest-musa -f llamacpp/musa.Dockerfile . 
```

```Dockerfile
COPY .git .git
COPY llamacpp/native/CMakeLists.txt .
COPY llamacpp/native/src src
COPY llamacpp/native/vendor vendor

# Fix submodule .git file to point to correct location in container
RUN echo "gitdir: ../../.git/modules/llamacpp/native/vendor/llama.cpp" > vendor/llama.cpp/.git && \
    sed -i 's|worktree = ../../../../../../llamacpp/native/vendor/llama.cpp|worktree = /llama-server/vendor/llama.cpp|' .git/modules/llamacpp/native/vendor/llama.cpp/config
```

But I don't know why to copy and fix the `.git` dir. Could you please explain? Thanks.

ericcurtin:
@p1-0tr or @doringeman can explain maybe

p1-0tr:
> But I don't know why to copy and fix the .git dir. Could you please explain? Thanks.

If this is not done, the `llama-server` build process won't be able to extract version info (tldr - the output of `llama-server --version` will be less useful)

ericcurtin:
If we can find a cleaner way, great

yeahdongcn:
> > But I don't know why to copy and fix the .git dir. Could you please explain? Thanks.
> 
> If this is not done, the `llama-server` build process won't be able to extract version info (tldr - the output of `llama-server --version` will be less useful)

If this is a must-have, should we update the paths to match the actual ones?

Please see the diff here: https://github.com/docker/model-runner/pull/299/commits/cbf0ac1422bfa22f1fc866f33e3e9483847eec29#diff-72094b40bd55a01f123ef8c0d4b78f1acf68cf12dbe784a38bdda5709158c9e4

PR #319, opened 2025-10-27
Provide clearer error message for unsupported platform in Install method

ilopezluna:
This pull request makes a minor improvement to the error message returned when the vLLM backend is not supported on the current platform. The new message is more descriptive, clarifying that the vLLM backend is not yet supported.


p1-0tr (approved):

ericcurtin:
I actually prefer the previous text (support in a loaded term, paid support?). Also the **_yet_** makes it sound like something is in progress (which often isn't the case).

PR #324, opened 2025-10-27
Use vllm official wheels

ilopezluna:
This pull request enhances the build of vllm docker image by using the prebuilt wheels of vllm


ericcurtin:
For x86_64 it looks like we want with +cu129? And for aarch64 without? assuming I interpreted things right from @simon-mo

PR #328, opened 2025-10-27
Automatically use ROCm when appropriate

ericcurtin:
On Linux at least

### DIFF
--- a/pkg/gpuinfo/amd_gpu_linux.go
+++ b/pkg/gpuinfo/amd_gpu_linux.go
@@ -1,0 +1,116 @@
+//go:build linux
+
+package gpuinfo
+
+import (
+	"bufio"
+	"os"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strconv"
+	"strings"
+)
+
+// supportedAMDGPUs are the AMD GPU targets that should use ROCm
+var supportedAMDGPUs = map[string]bool{
+	"gfx908":  true,
+	"gfx90a":  true,
+	"gfx942":  true,
+	"gfx1010": true,
+	"gfx1030": true,
+	"gfx1100": true,
+	"gfx1200": true,
+	"gfx1201": true,
+	"gfx1151": true,
+}
+
+func hasSupportedAMDGPU() (bool, error) {
+	// Check if KFD topology directory exists
+	topologyDir := "/sys/class/kfd/kfd/topology/nodes/"
+	info, err := os.Stat(topologyDir)
+	if err != nil || !info.IsDir() {
+		return false, nil // KFD not available
+	}
+
+	entries, err := os.ReadDir(topologyDir)
+	if err != nil {
+		return false, err
+	}
+
+	// Sort entries by name to maintain consistent order
+	sort.Slice(entries, func(i, j int) bool {
+		return entries[i].Name() < entries[j].Name()
+	})
+
+	// Compile regex to match gfx_target_version lines
+	reTarget := regexp.MustCompile(`gfx_target_version[ \t]+([0-9]+)`)
+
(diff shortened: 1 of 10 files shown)


pkg/inference/backends/llamacpp/download_linux.go
```diff
@@ -1,18 +1,50 @@
...
 )
 
-func (l *llamaCpp) ensureLatestLlamaCpp(_ context.Context, log logging.Logger, _ *http.Client,
-	_, vendoredServerStoragePath string,
+func init() {
+	// Enable GPU variant detection by default on Linux
+	ShouldUseGPUVariantLock.Lock()
+	defer ShouldUseGPUVariantLock.Unlock()
+	ShouldUseGPUVariant = true
+}
```
doringeman, line 19: I'd rather have `ShouldUseGPUVariant       = runtime.GOOS == "linux"` in `download.go` instead of an `init`.
For more context, this variable is used by Docker Desktop to determine whether it should display the GPU backed engine checkbox in the UI.

pkg/gpuinfo/gpuinfo_not_linux.go
```diff
@@ -0,0 +1,8 @@
+//go:build !linux
+
+package gpuinfo
+
+func (g *GPUInfo) HasSupportedAMDGPU() (bool, error) {
+	// AMD GPU detection is only supported on Linux
+	return false, nil
+}
```
doringeman, line 8:
```suggestion
}

```

pkg/gpuinfo/gpuinfo_linux.go
```diff
@@ -0,0 +1,3 @@
+//go:build linux
+
+package gpuinfo
```
doringeman, line 3:
```suggestion
package gpuinfo

```

pkg/gpuinfo/amd_gpu_linux.go
```diff
@@ -0,0 +1,116 @@
...
+			if supportedAMDGPUs[gfx] {
+				f.Close()
+				return true, nil // Found a supported AMD GPU
+			}
+		}
+		f.Close()
+	}
+
+	return false, nil // No supported AMD GPU found
+}
```
doringeman, line 116:
```suggestion
}

```

pkg/inference/backends/llamacpp/download_linux.go
```diff
@@ -1,18 +1,50 @@
...
+
+func (l *llamaCpp) ensureLatestLlamaCpp(ctx context.Context, log logging.Logger, httpClient *http.Client,
+	llamaCppPath, vendoredServerStoragePath string,
 ) error {
-	l.status = fmt.Sprintf("running llama.cpp version: %s",
-		getLlamaCppVersion(log, filepath.Join(vendoredServerStoragePath, "com.docker.llama-server")))
-	return errLlamaCppUpdateDisabled
```
p1-0tr, line 17: I think it'd be better to stick to a consistent logic for picking whether we use acceleration or not. Currently on LInux we expect people to run the model-runner container (appropriate to their usecase and system), so the accelerate/not-accelerate question should be decided in the CLI rather than here.

yeahdongcn:
Would you like me to add logic to auto-detect `musa` as well?

ericcurtin:
> Would you like me to add logic to auto-detect `musa` as well?

Sure, see what we did for Nvidia. But only if it's not crazy or unreliable. Ideally we should be able to probe /proc, /sys or /dev to figure this out (I know that is not always possible though)

yeahdongcn:
@ericcurtin Please see: https://github.com/docker/model-runner/pull/342

I will rebase main once your PR is merged.

PR #329, opened 2025-10-28
Install vllm

ilopezluna:
This pull request adds support for a new `--vllm` flag to the Docker Model Runner CLI commands (`install-runner`, `reinstall-runner`, and `start-runner`). The flag enables the use of the vLLM backend, but only when CUDA GPU support is available.

### DIFF
--- a/cmd/cli/commands/install-runner.go
+++ b/cmd/cli/commands/install-runner.go
@@ -116,7 +116,7 @@ func ensureStandaloneRunnerAvailable(ctx context.Context, printer standalone.Sta
 	}
 
 	// Ensure that we have an up-to-date copy of the image.
-	if err := standalone.EnsureControllerImage(ctx, dockerClient, gpu, false, printer); err != nil {
+	if err := standalone.EnsureControllerImage(ctx, dockerClient, gpu, "", printer); err != nil {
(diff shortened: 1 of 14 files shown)


ericcurtin:
Thoughts on "--backend vllm" ? In case we decide we want to implement another backend in future

ilopezluna:
> Thoughts on "--backend vllm" ? In case we decide we want to implement another backend in future

I like it! let me update it

ilopezluna:
@ericcurtin done in [8ec406e](https://github.com/docker/model-runner/pull/329/commits/8ec406eda8a1122303ee0e9224534d4cfbfe0bb7)

cmd/cli/commands/install-runner.go
```diff
@@ -254,9 +255,19 @@ func runInstallOrStart(cmd *cobra.Command, opts runnerOptions) error {
 		return fmt.Errorf("unknown GPU specification: %q", opts.gpuMode)
 	}
 
+	// Validate backend selection
+	if opts.backend != "" && opts.backend != "llama.cpp" && opts.backend != "vllm" {
```
doringeman, line 259: nit: use `llamacpp.Name` and `vllm.Name`.

PR #331, opened 2025-10-28
feat(cli): show context size in docker model ls

### LINKED ISSUE #246: Make `docker model ls` show `context_size` in its default output
In the pull request #243, a new way to repackage models with config-only changes (e.g. changing context size) its been introduced. 

However, after this change, viewing the context size of a model still requires doing:
`docker model inspect <model>`

One of the reviewers suggested:
“It will be good if the `docker model ls` can output `context_size` directly to avoid having to carry out an inspection for this parameter by `docker model inspect`.” 

That is a great usability improvement idea. I’m opening this issue to track and discuss adding that enhancement.

## Proposal / Use Cases

- **Quick glance**: When listing models, seeing context size immediately in `docker model ls` helps users identify which model variants have larger windows, especially when you have many variants.
- **Filtering / scripting**: It allows automation (scripts, CI tools) to pick models by context size without extra inspect calls.
- **Consistency**: The variant’s `context_size` is now a first-class parameter, it makes sense for it to be visible in summary listings.

For example:

```
REPOSITORY        SIZE     CMD       CTX
ai/smollm2        256 MiB  llama     2048
ai/othermodel     512 MiB  llama     4096
```

If a variant doesn’t set a context explicitly show default (4096).

jwjohns:
- Add a CONTEXT column to docker model ls so it’s easy to see each model’s window size.
  - Show the value from `config.context_size` when it exists, and fall back to `gguf["llama.context_length"]` for older models.
  - Right-align the new column so the table stays neat.

  ### Testing

  - `GOCACHE=/tmp/go-cache GOWORK=off go test ./...`
  - `docker model pull ai/smolvlm && make -C cmd/cli build && docker model ls`

  Closes #246.

### DIFF
--- a/cmd/cli/commands/list.go
+++ b/cmd/cli/commands/list.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -118,7 +119,7 @@ func prettyPrintModels(models []dmrm.Model) string {
 	var buf bytes.Buffer
 	table := tablewriter.NewWriter(&buf)
 
-	table.SetHeader([]string{"MODEL NAME", "PARAMETERS", "QUANTIZATION", "ARCHITECTURE", "MODEL ID", "CREATED", "SIZE"})
+	table.SetHeader([]string{"MODEL NAME", "PARAMETERS", "QUANTIZATION", "ARCHITECTURE", "MODEL ID", "CREATED", "CONTEXT", "SIZE"})
 
 	table.SetBorder(false)
 	table.SetColumnSeparator("")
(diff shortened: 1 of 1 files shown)


ericcurtin:
Thanks for the PR @jwjohns if the bot comment makes sense feel free to address in another PR, merging

ericcurtin (approved):

doringeman:
Hey @jwjohns, thank you for your contribution! Good catch! 😁 

I have the following notes about this PR:
- If there's no context size set, display an empty string to follow the docker ps format.
I.e.,
```
$ docker ps
CONTAINER ID   IMAGE     COMMAND     CREATED         STATUS        PORTS     NAMES
ac8d8d821d0f   alpine    "/bin/sh"   2 seconds ago   Up 1 second             admiring_lamport
```
- It doesn't display the [context size set by Compose/docker model configure](https://github.com/docker/compose/blob/main/pkg/compose/model.go#L156) - I believe this is important as might confuse users who don't use package for setting the context size.
- For llama.cpp models, should we display the default value? That's 4096, see [GetContextSize](https://github.com/docker/model-runner/blob/3f25b6584871ce54ba7df94232be2bcdcf5e4005/pkg/inference/backends/llamacpp/llamacpp_config.go#L81). This might be a bit more complicated to get, so don't bother too much for now.
- I believe we should also display the context size in `docker model ps`.

Would you like to address these in a follow up PR? Otherwise I would open an Issue to track them.

jwjohns:
@doringeman yes I was working on the code review fix currently. Shouldn't be long and I will have that fix in a PR for you guys.

aleksmelnikov:
I confirm, I do not see the stated changes.

```
wget https://github.com/docker/model-runner/archive/refs/tags/cmd/cli/v0.1.46.tar.gz
tar xvf v0.1.46.tar.gz
cd model-runner-cmd-cli-v0.1.46/
make -C cmd/cli build
 ./cmd/cli/model-cli version
Docker Model Runner version dev
Docker Engine Kind: Docker Engine
./cmd/cli/model-cli list
MODEL NAME             PARAMETERS  QUANTIZATION    ARCHITECTURE  MODEL ID      CREATED       SIZE
gemma3-qat:12B-Q4_K_M  11.77 B     IQ2_XXS/Q4_K_M  gemma3        701e72d4fb75  3 months ago  6.92 GiB
gemma3                 3.88 B      MOSTLY_Q4_K_M   gemma3        a353a8898c9d  5 weeks ago   2.31 GiB
```

doringeman:
Hey @aleksmelnikov, yes, because we haven't yet released it.
We plan to do it soon - early next week! 😁

PR #333, opened 2025-10-28
Fix/store consistency

### LINKED ISSUE #267: Failures between writing config, manifest, or tags may leave the store inconsistent
Related to [this comment](https://github.com/docker/model-runner/pull/243#discussion_r2444783956).

Currently, the methods that update the Store (e.g., writing the config, manifest, and tags) are not atomic.
If an error occurs in between these steps, for example, after writing the config but before the manifest or tags are persisted, the store may be left in an inconsistent or partially written state.

This can result in:

- Dangling references or incomplete entries in the store.
- Subsequent operations (reads, pushes, or deletions) failing unexpectedly.
- Difficulty in recovering or cleaning up after a failed write.

### Proposed solution

- Make the update operation atomic, either all store updates succeed, or none are persisted.
- Alternatively, implement a rollback mechanism to clean up any partially written data if an intermediate step fails.
- Add tests simulating mid-operation failures to verify that the store remains consistent after recovery.

jwjohns:
- make the store resilient to partial failures when writing models: the config, layers, manifest, and index now land only on success.
  - the new helper writes files through a temp-and-rename flow so updates are atomic, and Write/WriteLightweight capture cleanup actions if something fails mid-flight.
  - add `TestWriteRollsBackOnTagFailure` to prove that a broken tagging step leaves no blobs, manifests, or index entries behind.

  ### Testing

  - `GOCACHE=/tmp/go-cache GOWORK=off go test ./... `(full suite)
  - `GOCACHE=/tmp/go-cache GOWORK=off go test ./pkg/distribution/internal/store -run TestWriteRollsBackOnTagFailure -v` (explicit regression to show rollback works)

  Closes #267.

### DIFF
--- a/pkg/distribution/internal/store/blobs.go
+++ b/pkg/distribution/internal/store/blobs.go
@@ -1,7 +1,6 @@
 package store
 
 import (
-	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -180,15 +179,17 @@ func (s *LocalStore) writeConfigFile(mdl v1.Image) (bool, error) {
 	if err != nil {
 		return false, fmt.Errorf("get digest: %w", err)
 	}
+	hasBlob, err := s.hasBlob(hash)
+	if err != nil {
+		return false, fmt.Errorf("check config existence: %w", err)
+	}
+	if hasBlob {
+		return false, nil
+	}
+
 	path, err := s.blobPath(hash)
 	if err != nil {
 		return false, fmt.Errorf("get blob path: %w", err)
-	}
-	if _, err := os.Stat(path); err == nil {
-		// Config already exists in the store.
-		return false, nil
-	} else if !errors.Is(err, os.ErrNotExist) {
-		return false, fmt.Errorf("stat config blob: %w", err)
 	}
 
 	rcf, err := mdl.RawConfigFile()
--- a/pkg/distribution/internal/store/manifests.go
+++ b/pkg/distribution/internal/store/manifests.go
@@ -47,9 +47,12 @@ func (s *LocalStore) WriteManifest(hash v1.Hash, raw []byte) error {
 	if err := s.writeIndex(idx.Add(newEntryForManifest(hash, manifest))); err != nil {
 		// Best effort rollback to avoid leaving an orphaned manifest on disk.
 		if removeErr := s.removeManifest(hash); removeErr != nil && !errors.Is(removeErr, os.ErrNotExist) {
-			return fmt.Errorf("writing models index: %w (also failed to remove manifest: %v)", err, removeErr)
+			return errors.Join(
+				fmt.Errorf("write models index: %w", err),
+				fmt.Errorf("rollback remove manifest %s: %w", hash, removeErr),
+			)
 		}
-		return fmt.Errorf("writing models index: %w", err)
+		return fmt.Errorf("write models index: %w", err)
 	}
 	return nil
 }
@@ -75,7 +78,7 @@ func (s *LocalStore) removeManifest(hash v1.Hash) error {
 // writeFile is a wrapper around os.WriteFile that creates any parent directories as needed.
 func writeFile(path string, data []byte) error {
 	dir := filepath.Dir(path)
-	if err := os.MkdirAll(dir, 0777); err != nil {
+	if err := os.MkdirAll(dir, 0o755); err != nil {
 		return fmt.Errorf("create parent directory %q: %w", dir, err)
 	}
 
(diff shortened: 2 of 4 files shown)


ericcurtin:
@ilopezluna WDYT? PTAL

jwjohns:
Let me know what you guys think. This one was a little more in depth and I created it as draft to enable proper discussion prior to any merges.

pkg/distribution/internal/store/blobs.go
```diff
@@ -78,29 +79,33 @@ type blob interface {
...
 	hasBlob, err := s.hasBlob(hash)
 	if err != nil {
-		return fmt.Errorf("check blob existence: %w", err)
+		return false, v1.Hash{}, fmt.Errorf("check blob existence: %w", err)
 	}
 	if hasBlob {
-		// todo: write something to the progress channel (we probably need to redo progress reporting a little bit)
-		return nil
+		// TODO: write something to the progress channel (we probably need to redo progress reporting a little bit)
+		return false, hash, nil
```
ilopezluna, line 94: I was wondering, why not just use the `error` to determine whether the blob was created? If `error` is `nil`, it would mean the blob was created. However, this assumption is incorrect when the blob already exists. In that case, there’s no error, but no new blob is added. Therefore, there’s no need to keep track of the blob for cleanup in case of a rollback.

Did I understand that correctly?

pkg/distribution/internal/store/blobs.go
```diff
@@ -169,19 +174,29 @@ func incompletePath(path string) string {
...
-func (s *LocalStore) writeConfigFile(mdl v1.Image) error {
+// writeConfigFile writes the model config JSON file to the blob store and reports whether the file was newly created.
+func (s *LocalStore) writeConfigFile(mdl v1.Image) (bool, error) {
 	hash, err := mdl.ConfigName()
 	if err != nil {
-		return fmt.Errorf("get digest: %w", err)
+		return false, fmt.Errorf("get digest: %w", err)
 	}
-	rcf, err := mdl.RawConfigFile()
+	path, err := s.blobPath(hash)
```
ilopezluna: Oh we missed that case thanks!
You could also use `s.hasBlob(hash)` here

pkg/distribution/internal/store/manifests.go
```diff
@@ -43,7 +44,14 @@ func (s *LocalStore) WriteManifest(hash v1.Hash, raw []byte) error {
 		return fmt.Errorf("reading models: %w", err)
 	}
 
-	return s.writeIndex(idx.Add(newEntryForManifest(hash, manifest)))
+	if err := s.writeIndex(idx.Add(newEntryForManifest(hash, manifest))); err != nil {
+		// Best effort rollback to avoid leaving an orphaned manifest on disk.
+		if removeErr := s.removeManifest(hash); removeErr != nil && !errors.Is(removeErr, os.ErrNotExist) {
+			return fmt.Errorf("writing models index: %w (also failed to remove manifest: %v)", err, removeErr)
+		}
```
ilopezluna, line 54:
```suggestion
if removeErr := s.removeManifest(hash); removeErr != nil && !errors.Is(removeErr, os.ErrNotExist) {
		return errors.Join(
			fmt.Errorf("write models index: %w", err),
			fmt.Errorf("rollback remove manifest %s: %w", hash, removeErr),
		)
	}```

File: pkg/distribution/internal/store/store_test.go [hunk 4]
```diff
@@ -382,6 +382,88 @@ func TestStoreAPI(t *testing.T) {
 	})
 }
 
+func TestWriteRollsBackOnTagFailure(t *testing.T) {
```
ilopezluna, line 387: Thanks for adding a test of tag failure, could you also add one for config failure and layer failure? 🙏 
Ideally we should also add test of failure with an already existing blob but we can do it in a follow up PR

ilopezluna (approved):
@jwjohns thank you so much for looking into this! I don’t see anything blocking in this PR, so I’ve already approved it. I added some comments, but they’re just nice-to-have improvements we can merge this as it is and address them in a follow-up PR.
Of course, if you’d prefer to address the comments in this same PR, that’s perfect too!

jwjohns:
@ilopezluna i will get them handled shortly.

jwjohns:
@ilopezluna changes are in. I will take out of draft status and set for review for you. I went ahead and did everything except the blob check. I will get that PR in as well soon.

pkg/distribution/internal/store/store.go
```diff
@@ -200,10 +201,41 @@ func (s *LocalStore) Version() string {
...
+
+	type cleanupFunc func() error
+	var cleanups []cleanupFunc
+	success := false
+	defer func() {
+		if success {
+			return
+		}
+		for i := len(cleanups) - 1; i >= 0; i-- {
+			if err := cleanups[i](); err != nil && !errors.Is(err, os.ErrNotExist) {
```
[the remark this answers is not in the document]
jwjohns: will handle these as well.

jwjohns:
@ilopezluna I got your requested changes in and also the ones sorcery-ai recommended. I can do the blob in another PR tomorrow

PR #334, opened 2025-10-28
Fix potential hang on model load if eviction also runs

krissetto:
Particularly noticeable on systems with unified memory where currently VRAM is recognized as 0MB, like on Strix Halo based machines, switching models without manually unloading can cause the load process, and the whole `docker model run ...` command as well, to hang.

What this PR does is continue the load loop upon evictions so that all state is recalculated and loading of a new runner can continue. The rest of changes are to reduce code duplication, add info/debug logs add some tests

Repro (on linux and Strix Halo machine):
- run `docker model run ai/gpt-oss:20B-F16`
- ask some question so the model loads and starts responding
- exit the command
- run `docker model run ai/gemma3-qat:27B-Q4_K_M`
- ask some question, the previous model gets evicted, and the command hangs with no model loading and no output

Screencast of issue:
https://github.com/user-attachments/assets/d265780b-f7e9-42d5-9648-7c29570d4c3d

### DIFF
--- a/pkg/inference/scheduling/loader.go
+++ b/pkg/inference/scheduling/loader.go
@@ -211,6 +211,27 @@ func (l *loader) broadcast() {
 	}
 }
 
+// formatMemorySize formats a memory size in bytes as a string.
+// Values of 0 or 1 are treated as sentinel values for "unknown" memory size.
+func formatMemorySize(bytes uint64) string {
+	if bytes <= 1 {
+		return "unknown"
+	}
+	return fmt.Sprintf("%d MB", bytes/1024/1024)
+}
+
+// freeRunnerSlot frees a runner slot and reclaims its memory.
+// The caller must hold the loader lock.
+func (l *loader) freeRunnerSlot(slot int, key runnerKey) {
+	l.slots[slot].terminate()
+	l.slots[slot] = nil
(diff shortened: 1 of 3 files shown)


pkg/inference/scheduling/loader.go
```diff
@@ -480,6 +508,13 @@ func (l *loader) load(ctx context.Context, backendName, modelID, modelRef string
 	for {
 		slot := -1
 		availableVRAM := l.availableMemory.VRAM
+		if runtime.GOOS == "windows" {
```
ericcurtin, line 511: Honestly, I think we should turn off this memory estimation stuff, it's messy as hell, it's just the nature of all the different kinds of hardware, software, OS, etc.

ericcurtin (approved):

pkg/inference/scheduling/loader.go
krissetto: also limits unified memory machines to just a single model for now, which can sometimes be a bit frustrating..but i know i'm a small part of the audience, i wonder if the DGX has similar issues with unified memory or if its reported correctly

ericcurtin: I think unified memory is the future DGX Spark is unified memory (of course any ARM machine is), etc. There's a performance advantage of not having to copy between VRAM and RAM.

PR #344, opened 2025-10-29
refactor(scheduler): deduplicate vLLM backend selection logic

doringeman:
Move safetensors format detection into shared selectBackendForModel helper. Ensures both handleOpenAIInference and Configure consistently auto-select vLLM for safetensors models.

To test this you can do the following.

```
make docker-run-vllm
```

```
cat > test-compose-models.yaml << 'EOF'
services:
  ai_runner:
    image: alpine
    models:
      - llm 
models:
  llm:
    model: aistaging/smollm2-vllm
    context_size: 2048
EOF
```

```
MODEL_RUNNER_HOST=http://localhost:8080 docker compose -f test-compose-models.yaml --progress plain up
```

You can see it in the logs:
```
time="2025-10-29T11:17:15Z" level=info msg="vLLM args: [serve /models/bundles/sha256/d3d696dba6a8f2c00d3c7f1862f5b11f85b1fe3577d8d911a9cc314ed4de2be8/model --uds inference-runner-0.sock --max-model-len 2048 --served-model-name sha256:d3d696dba6a8f2c00d3c7f1862f5b11f85b1fe3577d8d911a9cc314ed4de2be8 aistaging/smollm2-vllm:latest]"
```

But also in the requests/responses recorder after you ran an inference request:

```
MODEL_RUNNER_HOST=http://localhost:8080 docker model run aistaging/smollm2-vllm hi
Hello there! I’m here to help with an interesting task. What can I assist you with?
```

```
MODEL_RUNNER_HOST=http://localhost:8080 docker model requests | jq .
[
  {
    "count": 1,
    "model": "sha256:d3d696dba6a8f2c00d3c7f1862f5b11f85b1fe3577d8d911a9cc314ed4de2be8",
    "config": {
      "context-size": 2048
    },
```

### DIFF
--- a/pkg/inference/scheduling/scheduler.go
+++ b/pkg/inference/scheduling/scheduler.go
@@ -155,6 +155,27 @@ func (s *Scheduler) Run(ctx context.Context) error {
 	return workers.Wait()
 }
 
+// selectBackendForModel selects the appropriate backend for a model based on its format.
+// If the model is in safetensors format, it will prefer vLLM if available.
+func (s *Scheduler) selectBackendForModel(model types.Model, backend inference.Backend, modelRef string) inference.Backend {
+	config, err := model.Config()
+	if err != nil {
+		s.log.Warnln("failed to fetch model config:", err)
+		return backend
+	}
+
+	if config.Format == types.FormatSafetensors {
+		if vllmBackend, ok := s.backends[vllm.Name]; ok && vllmBackend != nil {
+			return vllmBackend
+		}
+		s.log.Warnf("Model %s is in safetensors format but vLLM backend is not available. "+
+			"Backend %s may not support this format and could fail at runtime.",
+			utils.SanitizeForLog(modelRef), backend.Name())
+	}
+
+	return backend
+}
+
 // handleOpenAIInference handles scheduling and responding to OpenAI inference
 // requests, including:
 // - POST <inference-prefix>/{backend}/v1/chat/completions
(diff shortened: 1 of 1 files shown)


pkg/inference/scheduling/scheduler.go
```diff
@@ -155,6 +155,27 @@ func (s *Scheduler) Run(ctx context.Context) error {
...
+// If the model is in safetensors format, it will prefer vLLM if available.
+func (s *Scheduler) selectBackendForModel(model types.Model, backend inference.Backend, modelRef string) inference.Backend {
+	config, err := model.Config()
+	if err != nil {
+		s.log.Warnln("failed to fetch model config:", err)
+		return backend
+	}
+
+	if config.Format == types.FormatSafetensors {
+		if vllmBackend, ok := s.backends[vllm.Name]; ok {
```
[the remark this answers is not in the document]
doringeman: Fair!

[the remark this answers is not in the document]
doringeman: Done in https://github.com/docker/model-runner/compare/e2356868b36016c4555a930bf9e00e183e8a018d..a6f258bac41b4e7dc93233c3b09648c5bc2a3ca6.

pkg/inference/scheduling/scheduler.go
```diff
@@ -155,6 +155,27 @@ func (s *Scheduler) Run(ctx context.Context) error {
 	return workers.Wait()
 }
 
+// selectBackendForModel selects the appropriate backend for a model based on its format.
+// If the model is in safetensors format, it will prefer vLLM if available.
+func (s *Scheduler) selectBackendForModel(model types.Model, backend inference.Backend, modelRef string) inference.Backend {
```
p1-0tr, line 160: Passing `backend` here seems somewhat counter intuitive.

doringeman: Do you suggest returning `""` if it fails to automatically identify the backend?
And use it like `if b := selectBackendForModel(...); b != "" { backend = b }`?

ilopezluna: In case of error we could just return `defaultBackend`

doringeman: We'd still need to pass that as a parameter, right? Because this function is called on `/engines/{backend}/v1` as well.

ilopezluna: oh you're right, I thought `defaultBackend` was just a constant, but its dynamically assigned when `scheduling.NewScheduler` is called.
I think I would prefer to have the default backend hardcoded, for things like `selectBackendForModel`.
But we can discuss/change it in another pr 👍

ilopezluna (approved):

ericcurtin (approved):

PR #348, opened 2025-10-30
feat(cli): add purge command

doringeman:
```
make docker-run
```

```
make -C cmd/cli install
```

Screenshot 2025-10-30 at 12 35 22

### DIFF
--- a/cmd/cli/commands/purge.go
+++ b/cmd/cli/commands/purge.go
@@ -1,0 +1,47 @@
+package commands
+
+import (
+	"fmt"
+
+	"github.com/docker/model-runner/cmd/cli/commands/completion"
+	"github.com/docker/model-runner/cmd/cli/desktop"
+	"github.com/spf13/cobra"
+)
+
+func newPurgeCmd() *cobra.Command {
+	var force bool
+
+	c := &cobra.Command{
+		Use:   "purge [OPTIONS]",
+		Short: "Remove all models",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if !force {
+				cmd.Println("WARNING! This will remove the entire models directory.")
+				cmd.Print("Are you sure you want to continue? [y/N] ")
+
+				var input string
+				_, err := fmt.Scanln(&input)
+				if err != nil && err.Error() != "unexpected newline" {
+					return err
+				}
+
+				if input != "y" && input != "Y" {
+					cmd.Println("Operation cancelled.")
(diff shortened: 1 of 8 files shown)


cmd/cli/commands/purge.go
```diff
@@ -0,0 +1,47 @@
...
+				var input string
+				_, err := fmt.Scanln(&input)
+				if err != nil && err.Error() != "unexpected newline" {
+					return err
+				}
+
+				if input != "y" && input != "Y" {
+					cmd.Println("Operation cancelled.")
+					return nil
+				}
```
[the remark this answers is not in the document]
doringeman: `fmt.Scanln`, just like more `fmt` functions, returns space-delimited strings. See [`fmt` documentation](https://pkg.go.dev/fmt#:~:text=Input%20processed%20by%20verbs%20is%20implicitly%20space%2Ddelimited%3A%20the%20implementation%20of%20every%20verb%20except%20%25c%20starts%20by%20discarding%20leading%20spaces%20from%20the%20remaining%20input%2C%20and%20the%20%25s%20verb%20(and%20%25v%20reading%20into%20a%20string)%20stops%20consuming%20input%20at%20the%20first%20space%20or%20newline%20character.) and look for
```
// convertString returns the string represented by the next input characters.
// The format of the input is determined by the verb.
func (s *ss) convertString(verb rune) (str string)
```
in `toolchain@v0.0.1-go1.25.3.darwin-arm64/src/fmt/scan.go`.

cmd/cli/desktop/desktop.go
```diff
@@ -762,6 +762,22 @@ func (c *Client) Requests(modelFilter string, streaming bool, includeExisting bo
 	return resp.Body, cancel, nil
 }
 
+func (c *Client) Purge() error {
+	purgePath := inference.ModelsPrefix + "/purge"
```
ericcurtin, line 766: We should try and change this or alias it to /api/delete before it gets usage:

https://github.com/ollama/ollama/blob/main/docs/api.md#delete-a-model

ericcurtin: I guess this is the /api/delete all so it's different

doringeman: Yeah, that `DELETE /api/delete` is actually:
```
"DELETE " + inference.ModelsPrefix + "/{name...}":                     m.handleDeleteModel,
```

Issue #349, opened 2025-10-30
Inconsistent units ("M", " MiB", "M", " M")


ericcurtin:
```
~$ docker model ls
MODEL NAME                                      PARAMETERS  QUANTIZATION    ARCHITECTURE  MODEL ID      CREATED       SIZE
hf.co/ggml-org/smolvlm-500m-instruct-gguf:q8_0  409M                        llama         982f124fa007  6 months ago  437M
hf.co/ggml-org/gemma-3-270m-gguf                268M                        gemma3        845da567e322  2 months ago  292M
smollm2                                         361.82 M    IQ2_XXS/Q4_K_M  llama         354bf30d0aa3  7 months ago  256.35 MiB
ericcurtin044/smollm2                           361.82 M    IQ2_XXS/Q4_K_M  llama         354bf30d0aa3  7 months ago  256.35 MiB
ignaciolopezluna020/embedding:dir-test          134.52 M    MOSTLY_Q2_K     llama         0a3c1f157564  10 days ago   82.41 MiB
embedding:dir-test                              134.52 M    MOSTLY_Q2_K     llama         4e6611fbb0e9  10 days ago   82.41 MiB
embedding                                       134.52 M    MOSTLY_Q2_K     llama         ae9c542496a3  10 days ago   82.41 MiB
ecurtin@LQ5W arm64 11:12:23 0
~$ docker model ls | awk '{print $6}'
MODEL
months
months
354bf30d0aa3
354bf30d0aa3
0a3c1f157564
4e6611fbb0e9
ae9c542496a3
~$ docker images
REPOSITORY                          TAG       IMAGE ID       CREATED       SIZE
alpine                              latest    4b7ce07002c6   3 weeks ago   13.3MB
rw4lll/openwebui-docker-extension   0.1.0     b7e59b9001cc   5 weeks ago   14.3MB
```

Makes it hard to use tools like awk, arrange things in columns, whitespace separated

krissetto:
there should be a docker lib for these unit formatting scenarios: https://github.com/docker/go-units

hwclass:
Hi! I'm new to this codebase but after @ericcurtin's encouragement, wanted to give this a shot. I've implemented a fix (which works on my machine :) ) for the inconsistent unit formatting issue. Would love to get feedback before opening a PR!

## Approach

I've standardized unit formatting by removing spaces between numbers and units for both GGUF and Safetensors models.

[...]

PR #352, opened 2025-10-30
Sort output of `docker model list` to make it more human readable

krissetto:
I did this real quick on my fork because I was going mad trying to read what models and variants I had, so I figured I'd open a PR in case you also like this.

Took a somewhat opinionated double sorting approach, so the list results alphabetically ordered by model name (without tag), and then also sorted by tags for each base model name

**Screenshots**

Before:

image

---

After:  

Screenshot From 2025-10-30 14-45-37

### DIFF
--- a/cmd/cli/commands/list.go
+++ b/cmd/cli/commands/list.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


doringeman (commented):
Thanks! I wonder whether UX-wise it would be nice to try to sort them the same way as the `docker images` is. WDYT?

krissetto:
That'd be by creation date IIRC, so here it'd be.. pull date? I'm not super sure about that because it'd still be hard to read if i don't pull my models in some specific order. Good to chat about this though! 

One thing: lets keep in mind the `docker cli` has difficulties in changing some things, especially outputs, because third parties came to rely on specific behavior (even though it's not supposed to be an api, it gets used as such).
In this project we still have the ability to focus on the user first before things like that become a problem

ericcurtin:
Somewhat related to this is:

https://github.com/docker/model-runner/issues/349

PR #354, opened 2025-10-30
Add integration tests using Testcontainers

ilopezluna:
This pull request introduces integration testing support for the CLI, enabling end-to-end validation of model registry and runner interactions using Docker containers.

model-cli (what its being tested) <----> DMR (Testcontainer) <----> OCI Registry (Testcontainer)
The registry is set as the default one [in here](https://github.com/docker/model-runner/pull/354/files#diff-b2fe598f18e51fb911df56c09760bcd09ae26e411f8fa25e36902324614bbdecR157).

We can easily add more tests, for other commands like `tag`, `configure`, etc, but it can be done in a followup PR.
I added a separated workflow because it needs to build the DMR image (the current `latest` does not include the env variable to customize default registry yet), once we publish a new `latest` we can move the workflow to the normal CI.

The PR also contains a fix for the `inspect` command. 

@mdelapenya could you please take a look? Let us know any recommendation you might have 🙏

### DIFF
--- a/cmd/cli/commands/integration_test.go
+++ b/cmd/cli/commands/integration_test.go
@@ -19,6 +19,7 @@ import (
 	"github.com/docker/model-runner/pkg/distribution/registry"
 	"github.com/stretchr/testify/require"
 	"github.com/testcontainers/testcontainers-go"
+	tc "github.com/testcontainers/testcontainers-go/modules/registry"
 	"github.com/testcontainers/testcontainers-go/network"
 	"github.com/testcontainers/testcontainers-go/wait"
 )
@@ -132,10 +133,7 @@ func setupTestEnv(t *testing.T) *testEnv {
 
 func ociRegistry(t *testing.T, ctx context.Context, net *testcontainers.DockerNetwork) string {
 	t.Log("Starting OCI registry container...")
-	ctr, err := testcontainers.Run(
-		ctx, "registry:3",
-		testcontainers.WithExposedPorts("5000/tcp"),
-		testcontainers.WithWaitStrategy(wait.ForHTTP("/v2/").WithPort("5000/tcp").WithStartupTimeout(30*time.Second)),
+	ctr, err := tc.Run(context.Background(), "registry:3",
 		network.WithNetwork([]string{"registry.local"}, net),
 	)
 	require.NoError(t, err)
--- a/cmd/cli/go.mod
+++ b/cmd/cli/go.mod
@@ -24,6 +24,7 @@ require (
 	github.com/spf13/pflag v1.0.9
 	github.com/stretchr/testify v1.11.1
 	github.com/testcontainers/testcontainers-go v0.39.0
+	github.com/testcontainers/testcontainers-go/modules/registry v0.39.0
 	go.opentelemetry.io/otel v1.37.0
 	go.uber.org/mock v0.5.0
 	golang.org/x/sync v0.17.0
--- a/cmd/cli/go.sum
+++ b/cmd/cli/go.sum
@@ -334,6 +334,8 @@ github.com/stretchr/testify v1.11.1 h1:7s2iGBzp5EwR7/aIZr8ao5+dra3wiQyKjjFuvgVKu
 github.com/stretchr/testify v1.11.1/go.mod h1:wZwfW3scLgRK+23gO65QZefKpKQRnfz6sD981Nm4B6U=
 github.com/testcontainers/testcontainers-go v0.39.0 h1:uCUJ5tA+fcxbFAB0uP3pIK3EJ2IjjDUHFSZ1H1UxAts=
 github.com/testcontainers/testcontainers-go v0.39.0/go.mod h1:qmHpkG7H5uPf/EvOORKvS6EuDkBUPE3zpVGaH9NL7f8=
+github.com/testcontainers/testcontainers-go/modules/registry v0.39.0 h1:Wq08A4G5o/OYb68xWVzVWSHrpckpYab4+5u+8T5UaYQ=
+github.com/testcontainers/testcontainers-go/modules/registry v0.39.0/go.mod h1:RIRXImSUJ5MYAiM8Hl39JdMD6pHsRsgfhgR+L22dhMk=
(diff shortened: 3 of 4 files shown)


cmd/cli/commands/integration_test.go
```diff
@@ -0,0 +1,258 @@
...
+			pullRef:           networkFQDN2, // registry.local:5000/ai/test-model:latest
+			expectedModelID:   modelID2,
+			expectedModelName: "ai/test-model:latest",
+		},
+		{
+			name:              "default org without tag (should default to :latest)",
+			pullRef:           "registry.local:5000/ai/test-model",
+			expectedModelID:   modelID2,
+			expectedModelName: "ai/test-model:latest",
+		},
```
ilopezluna: Add cases using digests

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,37 @@
+name: Integration Tests
```
ilopezluna, line 1: I'm adding this additional workflow because I have to build the DMR image before running tests. This temporary, integration test should be run in the normal CI workflow

cmd/cli/commands/integration_test.go
```diff
@@ -0,0 +1,452 @@
...
+}
+
+// modelInfo contains all the information needed to generate model references
+type modelInfo struct {
+	name         string // e.g., "test-model"
+	org          string // e.g., "ai"
+	tag          string // e.g., "latest"
+	registry     string // e.g., "registry.local:5000"
+	modelID      string // Full ID: sha256:...
+	digest       string // sha256:...
```
ilopezluna, line 41: Note: `modelID` and `digest` are exactly the same (the digest of the manifest) I just added 2 fields to make it more readable

ilopezluna:
> LGTM! Why isn't the `integration-tests` rule in the root Makefile?

I keep it in a separated workflow for now because we need to build DMR image with current version of DMR, otherwise we cannot define the default registry

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,37 @@
...
+        run: make docker-build
+        env:
+          DOCKER_BUILDKIT: 1
+      
+      - name: Verify Docker image
+        run: docker images | grep docker/model-runner
+      
+      - name: Run integration tests
+        working-directory: cmd/cli/commands
+        run: go test -v -tags=integration -timeout=5m -run TestIntegration_PullModel
```
doringeman: Why only `PullModel`?

cmd/cli/commands/integration_test.go
```diff
@@ -0,0 +1,452 @@
...
+	return &testEnv{
+		ctx:         ctx,
+		registryURL: registryURL,
+		client:      client,
+	}
+}
+
+func ociRegistry(t *testing.T, ctx context.Context, net *testcontainers.DockerNetwork) string {
+	t.Log("Starting OCI registry container...")
+	ctr, err := testcontainers.Run(
```
mdelapenya: suggestion: use the registry module: https://golang.testcontainers.org/modules/registry/

cmd/cli/commands/integration_test.go
```diff
@@ -0,0 +1,452 @@
...
+	testcontainers.CleanupContainer(t, ctr)
+
+	registryEndpoint, err := ctr.Endpoint(ctx, "")
+	require.NoError(t, err)
+	registryURL := fmt.Sprintf("http://%s", registryEndpoint)
+	t.Logf("Registry available at: %s", registryURL)
+	return registryURL
+}
+
+func dockerModelRunner(t *testing.T, ctx context.Context, net *testcontainers.DockerNetwork) string {
```
mdelapenya, line 149: suggestion: use the DMR module: https://golang.testcontainers.org/modules/dockermodelrunner/

ilopezluna: @mdelapenya thanks, I tried at first but I think it can not help to our purpose because it doesn't spin up a a container with DMR, instead it relies on DMR of DD: https://github.com/testcontainers/testcontainers-go/blob/75272033cf089a08dbeaeb65cb9b312c7fadd557/modules/dockermodelrunner/dockermodelrunner.go#L15

mdelapenya: Oh I see, you need the in-code DMR, not the one installed in DD. Thanks for the clarification!

cmd/cli/commands/integration_test.go
```diff
@@ -0,0 +1,464 @@
+//go:build integration
```
mdelapenya, line 1: suggestion: I'd avoid using the build tag unless necessary. Why? Because then you'll be running tests twice (adding the flag in the make file will run all AND will include files with this tag), so you can end up adding exclusions and complex boolean logic for this. My recommendation is to run them altogether, and treat in-process integration tests as unit tests, as they are blazing fast thanks to containers.

cmd/cli/commands/integration_test.go
ilopezluna: done in [70ff06b](https://github.com/docker/model-runner/pull/354/commits/70ff06bf67cd9c788988e787491c31f90fccefb4)

cmd/cli/commands/integration_test.go
ilopezluna: That's a good point, thank you!

ericcurtin:
Minor nit, logs report error when we have a valid fallback:

```
cmd/cli/model-cli inspect smollm2:latest
INFO[0001] Getting model by reference: smollm2:latest    component=model-manager
ERRO[0001] Failed to get model: model not found reference: smollm2:latest  component=model-manager
INFO[0001] Getting model by reference: ai/smollm2:latest  component=model-manager
{
    "id": "sha256:354bf30d0aa3af413d2aa5ae4f23c66d78980072d1e07a5b0d776e9606a2f0b9",
    "tags": [
        "index.docker.io/ai/smollm2:latest",
        "my-smollm2:latest"
    ],
    "created": 1742816981,
    "config": {
        "format": "gguf",
        "quantization": "IQ2_XXS/Q4_K_M",
        "parameters": "361.82 M",
        "architecture": "llama",
        "size": "256.35 MiB"
    }
}
+ MODEL_RUNNER_HOST=http://localhost:13434
+ cmd/cli/model-cli inspect my-smollm2
```

ericcurtin (approved):

PR #356, opened 2025-10-31
build: add LOCAL_LLAMA for make run to run with locally build llama.cpp on macOS

doringeman:
```
$ MODEL_RUNNER_PORT=8080 make run LOCAL_LLAMA=1
CGO_ENABLED=1 go build -ldflags="-s -w" -o model-runner ./main.go
Using local llama.cpp build from llamacpp/install/bin
INFO[0000] Running on system with 10922 MB VRAM
INFO[0000] Running on system with 16384 MB RAM
INFO[0000] Successfully initialized store                component=model-manager
INFO[0000] LLAMA_SERVER_PATH: /Users/doringeman/workspace/model-runner/llamacpp/install/bin
INFO[0000] Metrics endpoint enabled at /metrics
INFO[0000] Listening on TCP port 8080
INFO[0000] downloadLatestLlamaCpp: latest, metal, /Users/doringeman/workspace/model-runner/llamacpp/install/bin, /Users/doringeman/workspace/model-runner/updated-inference/bin/com.docker.llama-server
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,8 +42,12 @@ model-distribution-tool:
 
 # Run the application locally
 run: build
-	LLAMA_ARGS="$(LLAMA_ARGS)" \
-	./$(APP_NAME)
+	@LLAMACPP_BIN="llamacpp/install/bin"; \
+	if [ "$(LOCAL_LLAMA)" = "1" ]; then \
+		echo "Using local llama.cpp build from $${LLAMACPP_BIN}"; \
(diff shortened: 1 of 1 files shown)


ericcurtin:
Somewhat related, I do think we could merge all the Linux Dockerfiles for llama.cpp and model runner into one, building and installing llama.cpp can be one step in an early layer with all the build deps:

```
COPY scripts/llamacpp-install.sh llamacpp-install.sh
RUN ./llamacpp-install.sh rocm
```

would de-duplicate a lot and make development easier...

PR #359, opened 2025-11-02
Remove duplicate splitArgs function

ericcurtin:
We have an exact duplicate in utils


ericcurtin:
@copilot fix this issue:

go: finding module for package github.com/docker/model-runner/pkg/utils
go: github.com/docker/model-runner imports
	github.com/docker/model-runner/pkg/utils: no matching versions for query "latest"

PR #366, opened 2025-11-03
fix(standalone): use StatusPrinter in tryGetBindAscendMounts

doringeman:
Replace `fmt` with `StatusPrinter`, print errors to stderr, and only show "NOT FOUND" messages when `DEBUG=1` is set.

Screenshot 2025-11-03 at 17 24 21

@leo-pony We don't print such logs for other bind mounts we attempt. WDYT?


leo-pony:
@doringeman I got. You can delete this log directly, or I can delete it if it's inconvenient for you.

PR #367, opened 2025-11-03
stripDefaultsFromModelName fixed to account for default registries

ericcurtin:
So we strip defaults fully.

### DIFF
--- a/cmd/cli/commands/utils.go
+++ b/cmd/cli/commands/utils.go
@@ -11,6 +11,7 @@ import (
 	"github.com/docker/model-runner/cmd/cli/desktop"
 	"github.com/docker/model-runner/cmd/cli/pkg/standalone"
 	"github.com/docker/model-runner/pkg/inference/backends/vllm"
+	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/moby/term"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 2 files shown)


cmd/cli/commands/utils.go
```diff
@@ -82,24 +82,46 @@ func asPrinter(cmd *cobra.Command) standalone.StatusPrinter {
...
 //   - "myorg/gemma3:latest" -> "myorg/gemma3"
 //   - "gemma3:latest" -> "gemma3"
+//   - "index.docker.io/ai/gemma3:latest" -> "gemma3"
+//   - "docker.io/ai/gemma3:latest" -> "gemma3"
+//   - "docker.io/myorg/gemma3:latest" -> "myorg/gemma3"
 //   - "hf.co/bartowski/model:latest" -> "hf.co/bartowski/model"
 func stripDefaultsFromModelName(model string) string {
-	// Check if model has ai/ prefix without tag (implicitly :latest) - strip just ai/
+	// First, let's handle the default registry prefixes
+	defaultRegistries := []string{"index.docker.io/", "docker.io/"}
```
ilopezluna, line 118: This is not true in case of [setting default registry](https://github.com/docker/model-runner/pull/358). I don't see any way to retrieve the default registry in ggcr so we would need to implement a `getDefaultRegistry` that checks if it has been set, and if not return `name.DefaultRegistry` and the alias

ericcurtin:
@ilopezluna ready for review

PR #368, opened 2025-11-03
Initialize termenv with color caching for interactive terminals

krissetto:
I noticed the interactive CLI would often behave weirdly and also spill OSC codes into the input prompt, in particular `11;rgb:0000/0000/0000R`

This happens when using `WithAutoStyle` in glamour because it queries the terminal to find out the background color. The fix here is to cache the term color via termenv at the very start of the function

---

**Screenshot**
image

### DIFF
--- a/cmd/cli/commands/run.go
+++ b/cmd/cli/commands/run.go
@@ -15,6 +15,7 @@ import (
 	"github.com/docker/model-runner/cmd/cli/commands/completion"
 	"github.com/docker/model-runner/cmd/cli/desktop"
 	"github.com/docker/model-runner/cmd/cli/readline"
+	"github.com/muesli/termenv"
 
 	"github.com/fatih/color"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


ilopezluna (approved):
Thanks!

cmd/cli/commands/run.go
```diff
@@ -715,6 +716,12 @@ func newRunCmd() *cobra.Command {
 
 			// Use enhanced readline-based interactive mode when terminal is available
 			if term.IsTerminal(int(os.Stdin.Fd())) {
+				// Initialize termenv with color caching before starting interactive session.
+				// This queries the terminal background color once and caches it, preventing
+				// OSC response sequences from appearing in stdin during the interactive loop.
+				termenv.SetDefaultOutput(
+					termenv.NewOutput(os.Stdout, termenv.WithColorCache(true)),
```
doringeman:
```suggestion
					termenv.NewOutput(asPrinter(cmd), termenv.WithColorCache(true)),
```
?

doringeman (approved):

cmd/cli/commands/run.go
krissetto: Also seems to work! I didn't notice that was being used, I pushed a quick change

doringeman: Yeah, it's just been added. 😅 Thanks a lot! Can I hit merge? 😁

PR #371, opened 2025-11-05
test: add integration tests for pushing models

ilopezluna:
This PR adds integration testing for push command. The model reference is not normalized on Push because it has to be already normalized on tagging

### DIFF
--- a/cmd/cli/commands/integration_test.go
+++ b/cmd/cli/commands/integration_test.go
@@ -637,7 +637,7 @@ func TestIntegration_PushModel(t *testing.T) {
 
 	// Set up custom registry with different alias
 	t.Log("Starting custom OCI registry container...")
-	customRegistryCtr, err := tc.Run(context.Background(), "registry:3",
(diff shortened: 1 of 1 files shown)


cmd/cli/commands/integration_test.go
```diff
@@ -627,3 +629,150 @@ func TestIntegration_TagModel(t *testing.T) {
...
 }
+
+// TestIntegration_PushModel tests pushing a model to registries with various reference formats
+// to ensure proper reference normalization and successful push operations.
+func TestIntegration_PushModel(t *testing.T) {
+	env := setupTestEnv(t)
+
+	// Set up custom registry with different alias
+	t.Log("Starting custom OCI registry container...")
+	customRegistryCtr, err := tc.Run(context.Background(), "registry:3",
```
doringeman:
```suggestion
	customRegistryCtr, err := tc.Run(t.Context(), "registry:3",
```

PR #375, opened 2025-11-05
revert: model name normalization

doringeman:
Temporary revert. Will reimplement in next release.

### DIFF
--- a/cmd/cli/commands/integration_test.go
+++ b/cmd/cli/commands/integration_test.go
@@ -52,14 +52,14 @@ type referenceTestCase struct {
 // generateReferenceTestCases generates a comprehensive list of reference formats for testing
 func generateReferenceTestCases(info modelInfo) []referenceTestCase {
 	cases := []referenceTestCase{
-		{
-			name: "short form (no registry, no org, no tag)",
-			ref:  info.name,
-		},
-		{
-			name: "with tag",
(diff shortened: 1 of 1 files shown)


ericcurtin:
Nack on this, what is the issue encountered?

ilopezluna:
> Nack on this, what is the issue encountered?

We found a few more cases where tagging and removal were failing.
Maybe, @doringeman, you could share the commands you ran if you still have them?

Our reasoning is that we’ve already spent a considerable amount of time trying to fix it without success. We all like the feature, but if we want to release the vLLM this week, we’ll need to revert it for now.

Once the vLLM is released and covered by tests, it’ll be easier to reintroduce the improvement. WDYT?

ericcurtin:
> > Nack on this, what is the issue encountered?
> 
> We found a few more cases where tagging and removal were failing. Maybe, @doringeman, you could share the commands you ran if you still have them?
> 
> Our reasoning is that we’ve already spent a considerable amount of time trying to fix it without success. We all like the feature, but if we want to release the vLLM this week, we’ll need to revert it for now.
> 
> Once the vLLM is released and covered by tests, it’ll be easier to reintroduce the improvement. WDYT?

Lets push forward I'm not sure any of these are significant, the only issue that has been communicated is this one:

```
docker model tag smollm2 smollm2
```

which this revert does not fix, it is pre-existing

Issue #395, opened 2025-11-10
When pulling ai/gpt-oss:latest it attempts to pull safetensors


ericcurtin:
**$ MODEL_RUNNER_HOST=http://localhost:13434 cmd/cli/model-cli pull ai/gpt-oss
INFO[44948] Pulling model: ai/gpt-oss:latest
INFO[44948] Starting model pull: ai/gpt-oss:latest        component=model-manager
WARN[44949] Unsupported model format for "ai/gpt-oss:latest": error while pulling model: safetensors models are not currently supported - this runner only supports GGUF format models
Error: Failed to pull model: pulling ai/gpt-oss failed with status 415 Unsupported Media Type: safetensors models are not currently supported - this runner only supports GGUF format models

Usage:
  model pull MODEL [flags]

Flags:
  -h, --help                          help for pull
      --ignore-runtime-memory-check   Do not block pull if estimated runtime memory for model exceeds system resources.

Failed to pull model: pulling ai/gpt-oss failed with status 415 Unsupported Media Type: safetensors models are not currently supported - this runner only supports GGUF format models**

doringeman:
WDYT of https://github.com/docker/model-runner/pull/396?

ilopezluna:
I've already added config data in `ai/gpt-oss`, so this specific case is already solved but https://github.com/docker/model-runner/pull/396 will help to not block users in case of other models without expected config

PR #396, opened 2025-11-10
fix(distribution): add warning for models with empty format field

doringeman:
Attempt to fix https://github.com/docker/model-runner/issues/395.

### DIFF
--- a/pkg/distribution/distribution/client.go
+++ b/pkg/distribution/distribution/client.go
@@ -145,7 +145,7 @@ func (c *Client) PullModel(ctx context.Context, reference string, progressWriter
 	}
 
 	// Check for supported type
-	if err := checkCompat(remoteModel); err != nil {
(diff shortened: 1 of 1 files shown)


ericcurtin:
We should avoid blocking the pull regardless, this is like the memory checking all over again.

Allow the user to pull regardless, vLLM is continually expanding support for new hardware and users may want to use tools like offload. Or they may want to push/pull models without running them.

ilopezluna:
> We should avoid blocking the pull regardless, this is like the memory checking all over again.
> 
> Allow the user to pull regardless, vLLM is continually expanding support for new hardware and users may want to use tools like offload. Or they may want to push/pull models without running them.

@ericcurtin if I understand correctly with this change we will not block the pull, right? we will just log the warning and we will continue pulling.

doringeman:
> users may want to use tools like offload

It works with Offload because this code runs on the backend side, hence in the docker/model-runner container.

PR #402, opened 2025-11-12
Add OCI annotations to layer descriptors for GGUF and SafeTensors

ericcurtin:
- Add annotation constants (filepath, metadata, mediatype.untested) to types package
- Add FileMetadata struct following OCI spec
- Update partial.NewLayer to populate layer annotations with file metadata
- Update ManifestForLayers to preserve annotations from our Layer type
- Add comprehensive tests for annotation presence and validity
- All layer types (GGUF, SafeTensors, License, MMProj, ChatTemplate) now include annotations

### DIFF
--- a/pkg/distribution/internal/gguf/model_test.go
+++ b/pkg/distribution/internal/gguf/model_test.go
@@ -2,6 +2,7 @@ package gguf_test
 
 import (
 	"encoding/json"
+	"path"
 	"path/filepath"
 	"testing"
 
@@ -108,8 +109,13 @@ func TestGGUF(t *testing.T) {
 			}
 
 			// Check for required annotation keys
-			if _, ok := layer.Annotations[types.AnnotationFilePath]; !ok {
+			filePath, ok := layer.Annotations[types.AnnotationFilePath]
+			if !ok {
 				t.Errorf("Expected annotation %s to be present", types.AnnotationFilePath)
+			}
+
+			if filePath != path.Base("dummy.gguf") {
+				t.Errorf("Expected file path annotation to be '%s', got '%s'", path.Base("dummy.gguf"), filePath)
 			}
 
 			if _, ok := layer.Annotations[types.AnnotationFileMetadata]; !ok {
@@ -135,6 +141,21 @@ func TestGGUF(t *testing.T) {
 			}
 			if metadata.Size == 0 {
 				t.Error("Expected file size to be non-zero")
+			}
+			if metadata.Typeflag != 0 {
+				t.Errorf("Expected Typeflag 0 for regular file, got %d", metadata.Typeflag)
+			}
+			if metadata.Mode == 0 {
+				t.Error("Expected file mode to be non-zero")
+			}
+			if metadata.ModTime.IsZero() {
+				t.Error("Expected modification time to be set")
+			}
+			if metadata.Uid != 0 {
+				t.Error("Expected Uid to be set with default 0")
+			}
+			if metadata.Gid != 0 {
+				t.Error("Expected Gid to be set with default 0")
 			}
 		})
 	})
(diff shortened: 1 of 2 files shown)


pkg/distribution/internal/gguf/model_test.go
```diff
@@ -91,6 +92,51 @@ func TestGGUF(t *testing.T) {
...
+				t.Fatalf("Expected 1 layer, got %d", len(manifest.Layers))
+			}
+
+			layer := manifest.Layers[0]
+			if layer.Annotations == nil {
+				t.Fatal("Expected annotations to be present")
+			}
+
+			// Check for required annotation keys
+			if _, ok := layer.Annotations[types.AnnotationFilePath]; !ok {
```
ilopezluna: Ensure the value of the annotation is the expected one

pkg/distribution/internal/gguf/model_test.go
```diff
@@ -91,6 +92,51 @@ func TestGGUF(t *testing.T) {
...
+			var metadata types.FileMetadata
+			if err := json.Unmarshal([]byte(metadataJSON), &metadata); err != nil {
+				t.Fatalf("Failed to unmarshal file metadata: %v", err)
+			}
+
+			// Verify metadata fields
+			if metadata.Name != "dummy.gguf" {
+				t.Errorf("Expected file name 'dummy.gguf', got '%s'", metadata.Name)
+			}
+			if metadata.Size == 0 {
```
ilopezluna, line 142: Include check for the other fields

pkg/distribution/internal/gguf/model_test.go
```diff
@@ -177,5 +223,49 @@ func TestGGUFShards(t *testing.T) {
 				t.Fatalf("Expected layer with media type %s, got %s", types.MediaTypeGGUF, manifest.Layers[0].MediaType)
 			}
 		})
+
+		t.Run("TestAnnotations", func(t *testing.T) {
```
ilopezluna, line 248: This seems like unnecessary. Is it testing anything that is not already tested by the previous one?
Actually we could remove the previous one and add more checks on this one to ensure the values of the annotations of each layer

pkg/distribution/internal/partial/layer.go
```diff
@@ -22,14 +26,49 @@ func NewLayer(path string, mt ggcrtypes.MediaType) (*Layer, error) {
...
+	fileInfo, err := f.Stat()
+	if err != nil {
+		return nil, err
+	}
+
+	// Create file metadata
+	metadata := types.FileMetadata{
+		Name:     filepath.Base(path),
+		Mode:     uint32(fileInfo.Mode().Perm()),
+		Uid:      0, // Default to 0 as os.FileInfo doesn't provide this on all platforms
```
ilopezluna, line 43: is it ok to use 0 here?

pkg/distribution/internal/partial/partial.go
```diff
@@ -157,11 +157,18 @@ func ManifestForLayers(i WithLayers) (*v1.Manifest, error) {
...
-		desc, err := partial.Descriptor(l)
-		if err != nil {
-			return nil, fmt.Errorf("get layer descriptor: %w", err)
+		// Check if this is our Layer type which embeds the full descriptor with annotations
+		if layer, ok := l.(*Layer); ok {
+			// Use the embedded descriptor directly to preserve annotations
+			layers = append(layers, layer.Descriptor)
+		} else {
+			// Fall back to partial.Descriptor for other layer types
+			desc, err := partial.Descriptor(l)
```
ilopezluna, line 166: I don't get if/else. Is it not always true?

ilopezluna: Ah this is because partial.Descriptor(l) will not include the annotations 👍

ilopezluna (approved):

PR #405, opened 2025-11-13
fix: increase installation wait tries from 20 to 60

ilopezluna:
With current tries I usually get: 
```
Status: Downloaded newer image for docker/model-runner:latest-vllm-cuda
Successfully pulled docker/model-runner:latest-vllm-cuda
Creating model storage volume docker-model-runner-models...
Starting model runner container docker-model-runner...
standalone model runner took too long to initialize
```
This PR increases the timeout to 30 seconds to give it more chances to successfully start


ilopezluna:
it fails because in Offload DMR is already running so it should be using reinstall instead.

PR #406, opened 2025-11-14
Temporarily removing NIM docs

ericcurtin:
It's not ready yet, will re-introduce when ready


ericcurtin:
@DC-Shi hope to re-introduce this

DC-Shi:
> @DC-Shi hope to re-introduce this

Got it, what kind of update should we made?

PR #410, opened 2025-11-14
Fix model name normalization in run command

krissetto:


ericcurtin:
We are trying our best to apply these daemon side, so we don't have to normalize for every client request. Funnily enough on this branch if I purge all models:

https://github.com/docker/model-runner/pull/413

and do

```
docker model run smollm2 hello
```

it works...

cmd/cli/commands/run_test.go
```diff
@@ -156,3 +157,147 @@ func TestRunCmdDetachFlag(t *testing.T) {
 		t.Errorf("Expected detach flag value to be true, got false")
 	}
 }
+
+// TestRunModelNameNormalization verifies that model names are normalized correctly
+// in the run command to ensure consistency with how models are stored after pulling
+func TestRunModelNameNormalization(t *testing.T) {
```
ilopezluna, line 163: This is equals to [this test](https://github.com/docker/model-runner/blob/63ef1612c1d696f9bc6355a429072e1b01b5746d/cmd/cli/commands/utils_test.go#L11), right?

cmd/cli/commands/run.go
```diff
@@ -676,6 +677,8 @@ func newRunCmd() *cobra.Command {
 				return nil
 			}
 
+			model = dmrm.NormalizeModelName(model)
```
ilopezluna, line 680: Why is this needed? Have you found an error?
In the backend we resolve the model name to the model ID for doing inference, so in case of error I think we should apply normalization in the API layer

krissetto: i think this pr is no longer needed since things have changed in the backend

ilopezluna: Ok! I'm closing it then, thanks!

PR #413, opened 2025-11-16
Start implementing /api

ericcurtin:
From ollama, this now works:

MODEL_RUNNER_PORT=13434 make run

OLLAMA_HOST="127.0.0.1:13434" ollama run ai/smollm2
OLLAMA_HOST="127.0.0.1:13434" ollama ls
OLLAMA_HOST="127.0.0.1:13434" ollama ps
OLLAMA_HOST="127.0.0.1:13434" ollama stop ai/smollm2
OLLAMA_HOST="127.0.0.1:13434" ollama pull ai/smollm2:135M-Q4_0

### DIFF
--- a/cmd/cli/commands/list.go
+++ b/cmd/cli/commands/list.go
@@ -47,7 +47,7 @@ func newListCmd() *cobra.Command {
 			if err != nil {
 				return err
 			}
-			cmd.Print(models)
+			fmt.Fprint(cmd.OutOrStdout(), models)
 			return nil
 		},
 		ValidArgsFunction: completion.ModelNamesAndTags(getDesktopClient, 1),
--- a/main.go
+++ b/main.go
@@ -20,6 +20,7 @@ import (
 	"github.com/docker/model-runner/pkg/inference/scheduling"
 	"github.com/docker/model-runner/pkg/metrics"
 	"github.com/docker/model-runner/pkg/middleware"
+	"github.com/docker/model-runner/pkg/ollama"
 	"github.com/docker/model-runner/pkg/routing"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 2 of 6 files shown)


cmd/cli/commands/list.go
```diff
@@ -47,7 +47,7 @@ func newListCmd() *cobra.Command {
 			if err != nil {
 				return err
 			}
-			cmd.Print(models)
+			fmt.Fprint(cmd.OutOrStdout(), models)
```
doringeman, line 50: I'd keep `cmd.Print(models)` for consistency with other files.

doringeman (commented):
I'm surprised by how much new code this needs.
I wonder if we can reduce some duplication?
For example, could `handleGetModels` call the newly added `GetModels`? Or we might even move more logic from the CLI to the backend to handle both clients (DMR and ollama) in one place. WDYT?

cmd/cli/commands/list.go
ericcurtin: Not sure about that, Print, prints to STDERR which doesn't make sense for output

ericcurtin:
De-duplicated getmodels code. Keep in mind we will likely end up deleting the custom apis as we migrate to the ollama-compatible APIs. De-duplicating could end up being wasted effort. But it will be a slow process as we need to migrate compose, UI, etc.

cmd/cli/commands/list.go
doringeman: Probably that's because you're not running it as a Docker CLI plugin (i.e., `./cmd/cli/model-cli ls 2>/dev/null`).

When it's running as a CLI plugin the output will be set to STDOUT.
Reference: https://github.com/docker/cli/blob/master/cli-plugins/plugin/plugin.go#L179.

```
$ make -C cmd/cli install
Building model-cli...
go build -ldflags="-s -w -X github.com/docker/model-runner/cmd/cli/desktop.Version=v1.0.3-13-g728eec15-dirty" -o model-cli .
Using existing binary model-cli
Linking model-cli to Docker CLI plugins directory...
Link created: /Users/doringeman/.docker/cli-plugins/docker-model

$ MODEL_RUNNER_HOST=http://localhost:8080 docker model ls 2>/dev/null
MODEL NAME                                      PARAMETERS  QUANTIZATION    ARCHITECTURE  MODEL ID      CREATED       CONTEXT  SIZE
...
```

PR #421, opened 2025-11-18
Warn on platforms that vLLM is not implemented on yet

ericcurtin:
Instead of blocking. This will help users distribute vLLM models and implement vLLM for alternate platforms.

### DIFF
--- a/pkg/distribution/distribution/client.go
+++ b/pkg/distribution/distribution/client.go
@@ -221,7 +221,7 @@ func (c *Client) PullModel(ctx context.Context, reference string, progressWriter
 	}
 
 	// Check for supported type
-	if err := checkCompat(remoteModel, c.log, reference); err != nil {
+	if err := checkCompat(remoteModel, c.log, reference, progressWriter); err != nil {
 		return err
 	}
 
@@ -489,7 +489,7 @@ func GetSupportedFormats() []types.Format {
 	return []types.Format{types.FormatGGUF}
 }
 
-func checkCompat(image types.ModelArtifact, log *logrus.Entry, reference string) error {
+func checkCompat(image types.ModelArtifact, log *logrus.Entry, reference string, progressWriter io.Writer) error {
 	manifest, err := image.Manifest()
 	if err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


pkg/distribution/distribution/errors.go
```diff
@@ -16,6 +16,6 @@ var (
 		"client supports only models of type %q and older - try upgrading",
 		types.MediaTypeModelConfigV01,
 	))
-	ErrUnsupportedFormat = errors.New("safetensors models are not currently supported - this runner only supports GGUF format models")
+	ErrUnsupportedFormat = errors.New("vLLM backend currently only implemented for x86_64 Nvidia platforms")
```
doringeman: This doesn't need to be an error anymore.
```suggestion
	ErrConflict = errors.New("resource conflict")
)

const warnUnsupportedFormat = "Model format 'Safetensors' not supported. " +
	"vLLM backend requires x86_64 Linux with NVIDIA GPU"
```

pkg/distribution/distribution/client.go
```diff
@@ -508,7 +508,12 @@ func checkCompat(image types.ModelArtifact, log *logrus.Entry, reference string)
 		log.Warnf("Model format field is empty for %s, unable to verify format compatibility",
 			utils.SanitizeForLog(reference))
 	} else if !slices.Contains(GetSupportedFormats(), config.Format) {
-		return ErrUnsupportedFormat
+		// Write warning but continue with pull
+		log.Warnf("vLLM backend currently only implemented for x86_64 Nvidia platforms")
+		if err := progress.WriteWarning(progressWriter, ErrUnsupportedFormat.Error()); err != nil {
```
doringeman:
```suggestion
		log.Warnf(warnUnsupportedFormat)
		if err := progress.WriteWarning(progressWriter, warnUnsupportedFormat); err != nil {
```

pkg/distribution/distribution/client_test.go
```diff
@@ -462,17 +462,23 @@ func TestClientPullModel(t *testing.T) {
...
+			// On non-Linux, verify that a warning was written
+			progressOutput := progressBuf.String()
+			if !strings.Contains(progressOutput, `"type":"warning"`) {
+				t.Fatalf("Expected warning message on non-Linux platforms, got output: %s", progressOutput)
 			}
-		} else {
-			// On non-Linux, should fail with ErrUnsupportedFormat
-			if !errors.Is(err, ErrUnsupportedFormat) {
-				t.Fatalf("Expected ErrUnsupportedFormat on non-Linux platforms, got: %v", err)
+			if !strings.Contains(progressOutput, "safetensors") {
```
doringeman, line 480:
```suggestion
			if !strings.Contains(progressOutput, warnUnsupportedFormat) {
```

ericcurtin:
I would pick up those nits, but they don't build as is

PR #430, opened 2025-11-21
feat: exclude git submodules from shellcheck validation

doringeman:
Inspired by https://github.com/docker/model-runner/pull/429#discussion_r2549430111.

Update validate target to dynamically read .gitmodules and exclude all submodule paths from shellcheck, in addition to the hardcoded pkg/go-containerregistry exclusion. Displays excluded paths for visibility.

This prevents shellcheck from running on vendored third-party code in submodules like llamacpp/native/vendor/llama.cpp.

```
$ make validate
find . -type f -name "*.sh" | grep -v "pkg/go-containerregistry\|llamacpp/native/vendor" | xargs shellcheck
✓ Shellcheck validation passed!
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,7 +82,8 @@ test-docker-ce-installation:
 	BASE_IMAGE=$(BASE_IMAGE) scripts/test-docker-ce-installation.sh
 
 validate:
-	find . -type f -name "*.sh" | grep -v pkg/go-containerregistry | xargs shellcheck
+	find . -type f -name "*.sh" | grep -v "pkg/go-containerregistry\|llamacpp/native/vendor" | xargs shellcheck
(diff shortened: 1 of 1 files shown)


ericcurtin:
This could be a lot simpler:

```
find . -type f -name "*.sh" | grep -v "pkg/go-containerregistry\|llamacpp/native/vendor" | xargs shellcheck
```

ericcurtin:
I hope we can get to a point where we remove both submodules:

```
llamacpp/native/vendor/nvapi # seems unnecessary and nvidia specific
llamacpp/native/vendor/llama.cpp # could become a shallow clone of a specific commit sha in a Dockerfule RUN stage when we finish unforking
```

Issue #432, opened 2025-11-21
Unable to import package github.com/docker/model-runner


doringeman:
After https://github.com/docker/model-runner/commit/6d1c75e39d36cecf7e0104697ea2e0ebf0bcf5e8:

```
go get github.com/docker/model-runner@v1.0.4

go: downloading github.com/docker/model-runner/pkg/go-containerregistry v0.0.0-00010101000000-000000000000
go: github.com/docker/model-runner imports
	github.com/docker/model-runner/pkg/inference/backends/llamacpp imports
	github.com/docker/model-runner/pkg/go-containerregistry/pkg/v1: github.com/docker/model-runner/pkg/go-containerregistry@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000
go: github.com/docker/model-runner imports
	github.com/docker/model-runner/pkg/metrics imports
	github.com/docker/model-runner/pkg/go-containerregistry/pkg/authn: github.com/docker/model-runner/pkg/go-containerregistry@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000
go: github.com/docker/model-runner imports
	github.com/docker/model-runner/pkg/metrics imports
	github.com/docker/model-runner/pkg/go-containerregistry/pkg/name: github.com/docker/model-runner/pkg/go-containerregistry@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000
go: github.com/docker/model-runner imports
	github.com/docker/model-runner/pkg/metrics imports
	github.com/docker/model-runner/pkg/go-containerregistry/pkg/v1/remote: github.com/docker/model-runner/pkg/go-containerregistry@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000
go: github.com/docker/model-runner imports
	github.com/docker/model-runner/pkg/inference/backends/llamacpp imports
	github.com/docker/model-runner/pkg/distribution/types imports
	github.com/docker/model-runner/pkg/go-containerregistry/pkg/v1/types: github.com/docker/model-runner/pkg/go-containerregistry@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000
go: github.com/docker/model-runner imports
	github.com/docker/model-runner/pkg/inference/models imports
	github.com/docker/model-runner/pkg/distribution/registry imports
[...]

PR #434, opened 2025-11-21
Fix go-containerregistry module path to resolve import issues

ericcurtin:
For projects that use this as a library


ericcurtin:
https://github.com/docker/model-runner/issues/432

PR #437, opened 2025-11-21
feat: wake up idle Docker Cloud contexts automatically

doringeman:
Detect and wake up idle Docker Cloud contexts during context detection.

---

```
make -C cmd/cli install
```

Before:
```
$ docker model ls
latest: Pulling from docker/model-runner
```

Now:
```
$ docker model ls
latest-cuda: Pulling from docker/model-runner
```

### DIFF
--- a/cmd/cli/desktop/context.go
+++ b/cmd/cli/desktop/context.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/containerd/errdefs"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/context/docker"
 	clientpkg "github.com/docker/docker/client"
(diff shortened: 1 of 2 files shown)


cmd/cli/desktop/context.go
```diff
@@ -131,8 +131,41 @@ func NewContextForTest(endpoint string, client DockerHttpClient) (*ModelRunnerCo
...
+
+	// Wake up Docker Cloud by triggering an empty ContainerCreate call.
+	dockerClient, err := DockerClientForContext(cli, cli.CurrentContext())
+	if err != nil {
+		return fmt.Errorf("failed to create Docker client: %w", err)
+	}
+
+	// The call is expected to fail, but it triggers Docker Cloud to wake up from idle.
+	_, err = dockerClient.ContainerCreate(ctx, nil, nil, nil, nil, "")
+	return nil
```
krissetto, line 183: should we verify something after this trigger has been sent? e.g. check if the label is no longer idle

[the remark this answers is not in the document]
doringeman: Good point, added in https://github.com/docker/model-runner/compare/e51a194db2a8f8af19f68950ea0444de76f04a4e..dc34c7525476588da58053c1b8219093a4a83e76.

doringeman: And return an error so it is displayed as a warning?

krissetto: If it remains idle, it may be worth erroring out. i was mostly wondering about what happens if the trigger doesn't actually wake up the idle cloud context. Kinda like a sanity check/confirmation before letting the cli proceed

doringeman: If we were to error out, we’d end up blocking something — and I’d rather avoid that for now. I’m more inclined to take a best-effort approach initially. If there are other Docker Cloud issues, they’ll surface as errors later anyway.

@ericcurtin WDYT?

ericcurtin: That sounds reasonable to me. We can always tighten this up and make it a hard error later if it turns out to hide too many issues. If you want to add an extra warning SGTM

PR #444, opened 2025-11-24
Use the latest vulkan packages from Ubuntu 26.04

ericcurtin:
Users have noticed some AI-related optimizations from 25.10, as an exception, build on 26.04 for the Vulkan variant of our image.


ericcurtin:
https://github.com/ggml-org/llama.cpp/pull/17439

doringeman:
Should llama.cpp also be bumped in this PR?

ericcurtin:
> Should llama.cpp also be bumped in this PR?

Makes no difference to this PR

PR #460, opened 2025-11-27
add X-Request-Origin header for Ollama completions

ilopezluna:
updates model usage tracking to ensure that requests proxied through Ollama endpoints are tagged with a specific header

### DIFF
--- a/pkg/ollama/handler.go
+++ b/pkg/ollama/handler.go
@@ -1,9 +1,11 @@
 package ollama
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


ericcurtin (approved):

pkg/inference/scheduling/scheduler.go
```diff
@@ -243,8 +243,21 @@ func (s *Scheduler) handleOpenAIInference(w http.ResponseWriter, r *http.Request
 			}
 			return
 		}
+		// Determine the action for tracking
+		action := "inference/" + backendMode.String()
```
doringeman, line 247: Can you move these after `selectBackendForModel`, please? 😬

ilopezluna: Absolutely. Just to be sure there is no functional difference right?
Asking because `s.tracker.TrackModel(model, r.UserAgent(), action)` is executed before `selectBackendForModel` in main

ilopezluna:
> Do you think we should do it for `configure` as well? See
> 
> https://github.com/docker/model-runner/blob/be2c61477f0642e1f3f482478d2d6cbb4b65b3ea/pkg/inference/scheduling/scheduler.go#L485
> 
> .

I think not for now. Currently Ollama handler uses the endpoint `configure` but this is going to change soon as we are going to use `Configure` from the service very soon. Then we can decide how do we want to track it.

doringeman (commented):
Tested with
```
DEBUG=1 MODEL_RUNNER_PORT=8080 make run LOCAL_LLAMA=1
```

On `OLLAMA_HOST=localhost:8080 ollama run ai/smollm2:latest hi` I can see in the logs `Tracked index.docker.io/ai/smollm2:latest latest with user agent: docker-model-runner ollama/completion`.

We reconstruct the request, hence we're missing ollama's UserAgent.

pkg/inference/scheduling/scheduler.go
doringeman: Yeah, that's a bug, it means we might record `llama.cpp` instead of `vLLM` before identifying it's actually the vLLM backend we want to use.

doringeman: Sorry, misread. Discussed on Slack.

PR #465, opened 2025-11-30
fix: Add conversational context to interactive CLI chat  

### LINKED ISSUE #351: Interactive chat in CLI through `run` doesn't send conversational context
It seems we only send individual messages, not the entire conversation list (which seems to work fine within the GUI):

```
kiview@KPDX439Y4Q dmr % docker model run ai/llama3.2
> What is 42?
The answer "42" is famously associated with Douglas Adams' science fiction series "The Hitchhiker's Guide to the Galaxy." In the series, a supercomputer named Deep Thought is asked to find the "Answer to the Ultimate Question of Life, the Universe, and Everything." After 7.5 million years of computation, Deep Thought reveals that the answer is indeed 42.

However, the characters in the series then realize that they don't actually know what the Ultimate Question is, making the answer essentially meaningless. The number 42 has since become a cultural reference point, symbolizing the search for meaning and the absurdity of the universe.

In a more mathematical sense, 42 is simply a number, and its properties or significance are largely dependent on the context in which it's used.
> And what is 43?
I don't have enough information to answer that question. You didn't provide a mathematical context or a problem to solve. Could you please provide more details or clarify what you're trying to calculate? I'll do my best to help.
> What did I ask you about?
You didn't ask me anything yet. This is the beginning of our conversation. What would you like to talk about? I can summarize our conversation at the end if you'd like.
> My name is Kevin
Hello Kevin! It's nice to meet you. Is there something I can help you with, or would you like to chat?
> What is my name?
I don't have any information about you, including your name. I'm a conversational AI, and our conversation just started. I don't retain any information about previous conversations or users. Each time you interact with me, it's a new conversation. Would you like to share your name with me?
```

We can see this also happening, by using the request inspector:
```
{
  "messages": [
    {
      "content": "What is my name?",
      "role": "user"
    }
  ],
  "model": "ai/llama3.2:latest",
  "stream": true
}
```

nithish-95:
Fixes #351                                                                                                                                                                 
 This PR addresses an issue where the interactive chat in the CLI was not sending the full conversational context with each message. This resulted in the model not being able to remember previous parts of the conversation.                                                                                                                       
                                                                                                                                                                           
The following changes have been made:                                                                                                                                      

- A `messages` slice is now maintained to store the history of the conversation.                                                                                           
- On each new user message, the entire conversation history is sent to the model.                                                                                          
- The assistant's response is then appended to the history for the next turn.                                                                                             

This ensures that the model has the full context of the conversation, as demonstrated in the updated behavior.

### DIFF
--- a/cmd/cli/commands/nim.go
+++ b/cmd/cli/commands/nim.go
@@ -1,3 +1,4 @@
+package commands
 
 import (
 	"bufio"
--- a/pkg/inference/backends/llamacpp/download.go
+++ b/pkg/inference/backends/llamacpp/download.go
@@ -27,7 +27,7 @@ const (
 var (
 	ShouldUseGPUVariant       bool
 	ShouldUseGPUVariantLock   sync.Mutex
-	ShouldUpdateServer        = true
+	ShouldUpdateServer        = false
 	ShouldUpdateServerLock    sync.Mutex
 	DesiredServerVersion      = "latest"
 	DesiredServerVersionLock  sync.Mutex
--- a/pkg/inference/backends/llamacpp/llamacpp.go
+++ b/pkg/inference/backends/llamacpp/llamacpp.go
@@ -95,8 +95,8 @@ func (l *llamaCpp) Install(ctx context.Context, httpClient *http.Client) error {
 
 	// We don't currently support this backend on Windows. We'll likely
 	// never support it on Intel Macs.
-	if (runtime.GOOS == "darwin" && runtime.GOARCH == "amd64") ||
-		(runtime.GOOS == "windows" && !(runtime.GOARCH == "amd64" || runtime.GOARCH == "arm64")) {
+	// We don't currently support this backend on Windows.
+	if (runtime.GOOS == "windows" && !(runtime.GOARCH == "amd64" || runtime.GOARCH == "arm64")) {
 		return errors.New("platform not supported")
 	}
 


pkg/inference/backends/llamacpp/download.go
```diff
@@ -28,7 +27,7 @@ const (
 var (
 	ShouldUseGPUVariant       bool
 	ShouldUseGPUVariantLock   sync.Mutex
-	ShouldUpdateServer        = true
+	ShouldUpdateServer        = false
```
ilopezluna, line 30: Why should we not update llama.cpp?

pkg/inference/backends/llamacpp/llamacpp.go
```diff
@@ -95,8 +95,8 @@ func (l *llamaCpp) Install(ctx context.Context, httpClient *http.Client) error {
 
 	// We don't currently support this backend on Windows. We'll likely
 	// never support it on Intel Macs.
-	if (runtime.GOOS == "darwin" && runtime.GOARCH == "amd64") ||
-		(runtime.GOOS == "windows" && runtime.GOARCH != "amd64" && runtime.GOARCH != "arm64") {
+	// We don't currently support this backend on Windows.
```
ilopezluna, line 98:
```suggestion
```

pkg/inference/backends/llamacpp/llamacpp.go
```diff
@@ -95,8 +95,8 @@ func (l *llamaCpp) Install(ctx context.Context, httpClient *http.Client) error {
...
+	if (runtime.GOOS == "windows" && !(runtime.GOARCH == "amd64" || runtime.GOARCH == "arm64")) {
```
ilopezluna, line 99: Is it Intel Mac supported? If so we should update the comment. If not we should get back the `if (runtime.GOOS == "darwin" && runtime.GOARCH == "amd64") ||`

cmd/cli/commands/nim.go
```diff
@@ -407,15 +409,25 @@ func chatWithNIM(cmd *cobra.Command, model, prompt string) error {
...
+
+	requestPayload := struct {
+		Model    string                      `json:"model"`
+		Messages []desktop.OpenAIChatMessage `json:"messages"`
+		Stream   bool                        `json:"stream"`
+	}{
+		Model:    modelName,
+		Messages: *messages,
+		Stream:   true,
+	}
```
ilopezluna, line 423:
```suggestion
	requestPayload := desktop.OpenAIChatRequest{
		Model:    modelName,
		Messages: *messages,
		Stream:   true,
	}
```

ilopezluna (commented):
Thanks for working on this @nithish-95 !
I have tried your PR and it works well, nice 👏 
Added a couple of minor suggestions, and a questions related to llamacppgo and download.go changes (I think we should not change these files). 
Let me know what do you think 🙏

pkg/inference/backends/llamacpp/llamacpp.go
ericcurtin: Yes duplicate

pkg/inference/backends/llamacpp/llamacpp.go
ericcurtin: Personally I'd vote for removing lines 96 to 101 completely. But this is fine.

When somebody encounters an error from a new platform, I think this is good. It would be nice to encourage people to explore these errors and even enable new platforms (from a community perspective).

I think we try too hard to block code execution for things deemed unsupported. If the community wants to work on Intel Mac and get that working, so be it.

Now whether than propagates to Docker Desktop, is a different discussion (and unlikely since it's a legacy platform). We have the same discussion for Linux DMR.

pkg/inference/backends/llamacpp/download.go
ericcurtin: Would be nice to have an explanation of why this needed to be changed alright. Maybe the PR is not ready for review?

ericcurtin:
I meant to press comment instead of approve :)

ilopezluna:
@nithish-95 could you address the comments? 🙏

ericcurtin:
Thanks @nithish-95 we caught this here:

https://github.com/docker/model-runner/pull/532

please feel free to open PRs again.

PR #482, opened 2025-12-05
Remove runtime flags

ilopezluna:
This pull request refactors backend configuration handling for the CLI and inference backends, removing the generic `runtime-flags` option in favor of backend-specific, structured configuration. It introduces new flags for vLLM (`--hf_overrides`) and llama.cpp (`--reasoning-budget`), updates documentation, and adds comprehensive tests for the new flags.

### DIFF
--- a/cmd/cli/go.mod
+++ b/cmd/cli/go.mod
@@ -88,6 +88,7 @@ require (
 	github.com/magiconair/properties v1.8.10 // indirect
 	github.com/mattn/go-colorable v0.1.13 // indirect
 	github.com/mattn/go-isatty v0.0.20 // indirect
+	github.com/mattn/go-shellwords v1.0.12 // indirect
 	github.com/microcosm-cc/bluemonday v1.0.27 // indirect
 	github.com/mitchellh/go-homedir v1.1.0 // indirect
 	github.com/moby/docker-image-spec v1.3.1 // indirect
--- a/cmd/cli/go.sum
+++ b/cmd/cli/go.sum
@@ -219,6 +219,8 @@ github.com/mattn/go-runewidth v0.0.9/go.mod h1:H031xJmbD/WCDINGzjvQ9THkh0rPKHF+m
 github.com/mattn/go-runewidth v0.0.12/go.mod h1:RAqKPSqVFrSLVXbA8x7dzmKdmGzieGRCM46jaSJTDAk=
 github.com/mattn/go-runewidth v0.0.16 h1:E5ScNMtiwvlvB5paMFdw9p4kSQzbXFikJ5SQO6TULQc=
 github.com/mattn/go-runewidth v0.0.16/go.mod h1:Jdepj2loyihRzMpdS35Xk/zdY8IAYHsh153qUoGf23w=
+github.com/mattn/go-shellwords v1.0.12 h1:M2zGm7EW6UQJvDeQxo4T51eKPurbeFbe8WtebGE2xrk=
+github.com/mattn/go-shellwords v1.0.12/go.mod h1:EZzvwXDESEeg03EKmM+RmDnNOPKG4lLtQsUlTZDWQ8Y=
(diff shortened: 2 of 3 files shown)


cmd/cli/commands/configure.go
```diff
@@ -15,32 +16,21 @@ func newConfigureCmd() *cobra.Command {
...
+			if len(args) != 1 {
+				return fmt.Errorf(
+					"Exactly one model must be specified, got %d: %v\n\n"+
+						"See 'docker model configure --help' for more information",
+					len(args), args)
 			}
 			opts.Model = args[0]
-			opts.RuntimeFlags = args[1:]
 			return nil
 		},
```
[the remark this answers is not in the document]
ilopezluna: `--embeddings` flag in CLI makes no sense

pkg/inference/scheduling/api.go
```diff
@@ -93,10 +93,12 @@ type UnloadResponse struct {
...
-	RawRuntimeFlags string                               `json:"raw-runtime-flags,omitempty"`
-	Speculative     *inference.SpeculativeDecodingConfig `json:"speculative,omitempty"`
+	Model       string                               `json:"model"`
+	ContextSize int64                                `json:"context-size,omitempty"`
+	Mode        *inference.BackendMode               `json:"mode,omitempty"`
+	Speculative *inference.SpeculativeDecodingConfig `json:"speculative,omitempty"`
+
+	// Backend-specific configuration
+	VLLM     *inference.VLLMConfig     `json:"vllm,omitempty"`
+	LlamaCpp *inference.LlamaCppConfig `json:"llamacpp,omitempty"`
```
doringeman:
```suggestion
	Model string                 `json:"model"`
	Mode  *inference.BackendMode `json:"mode,omitempty"`
	inference.BackendConfiguration
```
WDYT of embedding instead?

IMO this pattern makes it immediately clear that ConfigureRequest embeds BackendConfiguration, which helps with code readability.
Also, when new fields are added to BackendConfiguration, if they need special default values, you initialize them in one place rather than hunting down all the places where ConfigureRequest is created.

cmd/cli/commands/configure.go
```diff
@@ -15,32 +16,21 @@ func newConfigureCmd() *cobra.Command {
 	var draftModel string
 	var numTokens int
 	var minAcceptanceRate float64
+	var hfOverrides string
+	var reasoningBudget int64
 
 	c := &cobra.Command{
-		Use:    "configure [--context-size=<n>] [--speculative-draft-model=<model>] MODEL [-- <runtime-flags...>]",
+		Use:    "configure [--context-size=<n>] [--speculative-draft-model=<model>] [--hf_overrides=<json>] [--reasoning-budget=<n>] MODEL",
```
ericcurtin, line 23: We should be careful about adding things like hf_overrides to our API, we are getting into the territory of inventing a DMR API again.

ilopezluna: @ericcurtin we need this specific vllm arg https://docs.vllm.ai/en/latest/configuration/model_resolution/ to support qwen3 rerank `--hf_overrides '{"architectures": ["Qwen3ForSequenceClassification"],"classifier_from_token": ["no", "yes"],"is_original_qwen3_reranker": true}'`

So I think our options are only:
1- do not support the model
2- include this specific arg

ericcurtin: Personally I would go 1. and open pull request to fix in vllm or open an issue in vllm of why we need these args (when a model is integrated well into vLLM or llama.cpp, it typically needs no args):

https://github.com/vllm-project/vllm/issue

Also, in llama.cpp we can detect and set this based on model metadata, why can't we do that here?

Keep in mind the users have a solution, use llama.cpp .

I think 2. should be close to a last resort if there is great user demand for this specific thing on vLLM.

But if you feel strongly about it, keep it in.

ilopezluna: That's fair, but because we supported this model before this PR I prefer to keep supporting it. Having said that, I'm going to open an issue in vllm about it. 
As soon as this arg is not required anymore I will remove it from DMR 👍

PR #484, opened 2025-12-05
chore: rely on the Go workspace

doringeman:
When another module tries to import `github.com/docker/model-runner/cmd/cli`, Go reads the `cmd/cli/go.mod` file and sees:

```
require github.com/docker/model-runner v0.0.0
replace github.com/docker/model-runner => ../..
```

Issues:
1. The v0.0.0 version doesn't exist - it's a placeholder
2. The replace directive uses a relative path (../..) that only works locally
3. External importers can't resolve the relative path - it points to a non-existent location on their filesystem
4. This makes cmd/cli effectively unimportable by external modules

### DIFF
--- a/cmd/cli/go.mod
+++ b/cmd/cli/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/docker/docker v28.3.3+incompatible
 	github.com/docker/go-connections v0.6.0
 	github.com/docker/go-units v0.5.0
-	github.com/docker/model-runner v0.0.0
+	github.com/docker/model-runner v1.0.3
 	github.com/docker/model-runner/pkg/go-containerregistry v0.0.0-20251121150728-6951a2a36575
 	github.com/emirpasic/gods/v2 v2.0.0-alpha
 	github.com/fatih/color v1.18.0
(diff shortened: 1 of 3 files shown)


cmd/cli/go.mod
```diff
@@ -70,6 +70,7 @@ require (
 	github.com/go-logr/stdr v1.2.2 // indirect
 	github.com/go-ole/go-ole v1.3.0 // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
+	github.com/google/go-containerregistry v0.20.6 // indirect
```
ilopezluna, line 73: don't we need to use the ggrc included in DMR?

ericcurtin: Yeah this will break resumable downloads

doringeman: It's used, it's referred as `github.com/docker/model-runner/pkg/go-containerregistry` where it needs to be.

ericcurtin: Ok as long as it doesn't break anything

doringeman: image

PR #495, opened 2025-12-09
Simplify configure

ilopezluna:
`--reasoning-budget N` can only be 0 or -1 from [here](https://github.com/ggml-org/llama.cpp/tree/master/tools/server):

> controls whether thought tags are allowed and/or extracted from the response, and in which format they're returned; one of:
> - none: leaves thoughts unparsed in message.content
> - deepseek: puts thoughts in message.reasoning_content
> - deepseek-legacy: keeps <think> tags in message.content while also populating message.reasoning_content
> (default: auto)
> (env: LLAMA_ARG_THINK)

So we can only support partially Ollama API: https://docs.ollama.com/api/generate#body-think-one-of-0 

In this PR I also rollback changes of `compose.go` as it's not used anymore, current compose integration uses `docker model configure` so I have updated that command instead.

### DIFF
--- a/cmd/cli/commands/configure_flags.go
+++ b/cmd/cli/commands/configure_flags.go
@@ -49,6 +49,41 @@ func (v *Int32PtrValue) Type() string {
 	return "int32"
 }
 
+// BoolPtrValue implements pflag.Value interface for *bool pointers
+// This allows flags to have a nil default value to detect if explicitly set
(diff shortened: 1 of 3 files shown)


ericcurtin:
Boolean thinking (partial Ollama API) support sounds fine to me, it's a small subset of people that will want to do this anyway

PR #514, opened 2025-12-15
Fix package command

ilopezluna:
In this PR, I’m fixing the `package` command by moving the normalization of the model name into the model distribution. This centralizes the normalization logic in a single place, making it less error-prone and easier to understand.

I’ve also updated the model filtering used by the `list` command so that it matches the behavior of the Docker CLI.

### DIFF
--- a/pkg/distribution/distribution/client.go
+++ b/pkg/distribution/distribution/client.go
@@ -580,7 +580,11 @@ func (c *Client) PushModel(ctx context.Context, tag string, progressWriter io.Wr
 // The layers must already exist in the store.
 func (c *Client) WriteLightweightModel(mdl types.ModelArtifact, tags []string) error {
 	c.log.Infoln("Writing lightweight model variant")
-	return c.store.WriteLightweight(mdl, tags)
+	normalizedTags := make([]string, len(tags))
+	for i, tag := range tags {
+		normalizedTags[i] = c.normalizeModelName(tag)
+	}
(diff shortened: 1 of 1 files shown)


cmd/cli/commands/integration_test.go
```diff
@@ -587,32 +587,6 @@ func TestIntegration_TagModel(t *testing.T) {
 		})
 	}
```
ilopezluna, line 589: Removed this verification because it's redundant. If a tag fails the individual test will fail, so this test is not adding any additional coverage and it becomes a false positive in some cases (for example when tagging a model using same reference as tag, its a no op, but this test will fail because the expected number of tagged models will not match with the actual)

cmd/cli/commands/list.go
```diff
@@ -72,27 +72,39 @@ func listModels(openai bool, desktopClient *desktop.Client, quiet bool, jsonForm
 	}
 
 	if modelFilter != "" {
-		// Normalize the filter to match stored model names (backend normalizes when storing)
-		normalizedFilter := dmrm.NormalizeModelName(modelFilter)
+		// If filter doesn't contain '/', prepend default namespace 'ai/'
```
ilopezluna, line 75: Updated the model filter logic to match the same logic as docker cli you can filter by repo + tag but not only by tag

doringeman (approved):

doringeman:
I'll take a closer look at while the lint fails on the CI, but not locally. Feel free to merge.

PR #516, opened 2025-12-15
feat(distribution): add CNCF ModelPack format compatibility

thc1006:
## Summary

Add support for pulling models in CNCF ModelPack format by converting them to Docker model-spec format on-the-fly.

- Add modelpack package with type definitions and conversion logic
- Convert config fields (paramSize -> parameters, createdAt -> created)
- Convert layer media types (cncf.model.weight -> docker.ai.gguf)
- Preserve extended metadata in Config.ModelPack extension field

## Related Issues

- https://github.com/docker/model-spec/issues/3
- https://github.com/modelpack/model-spec/issues/151

## Test Plan

- Unit tests for config conversion
- Unit tests for layer media type conversion
- Unit tests for artifact wrapper

### DIFF
--- a/cmd/cli/commands/list.go
+++ b/cmd/cli/commands/list.go
@@ -14,6 +14,7 @@ import (
 	"github.com/docker/model-runner/cmd/cli/commands/formatter"
 	"github.com/docker/model-runner/cmd/cli/desktop"
 	"github.com/docker/model-runner/cmd/cli/pkg/standalone"
+	"github.com/docker/model-runner/pkg/distribution/types"
 	dmrm "github.com/docker/model-runner/pkg/inference/models"
 	"github.com/olekukonko/tablewriter"
 	"github.com/spf13/cobra"
@@ -258,10 +259,10 @@ func appendRow(table *tablewriter.Table, tag string, model dmrm.Model) {
 	// Strip default "ai/" prefix and ":latest" tag for display
 	displayTag := stripDefaultsFromModelName(tag)
 	contextSize := ""
-	if model.Config.ContextSize != nil {
-		contextSize = fmt.Sprintf("%d", *model.Config.ContextSize)
-	} else if model.Config.GGUF != nil {
-		if v, ok := model.Config.GGUF["llama.context_length"]; ok {
+	if model.Config.GetContextSize() != nil {
+		contextSize = fmt.Sprintf("%d", *model.Config.GetContextSize())
+	} else if dockerConfig, ok := model.Config.(*types.Config); ok && dockerConfig.GGUF != nil {
+		if v, ok := dockerConfig.GGUF["llama.context_length"]; ok {
 			if parsed, err := strconv.ParseUint(v, 10, 64); err == nil {
 				contextSize = fmt.Sprintf("%d", parsed)
 			} else {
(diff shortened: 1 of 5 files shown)


ericcurtin:
What do you think about the AI review bot comments? What do you think about converting vs Docker pulling this format as is without conversion, so it can push/pull/run without conversion? Pros/cons?

thc1006:
@ericcurtin 

Thank you very much for your thoughtful review and for raising this important architectural question. I truly appreciate you taking the time to guide me on this.

---

### Regarding AI Review Bot Comments

I've reviewed all the suggestions carefully. They are helpful improvements (deep copy, test coverage, etc.). I'm happy to address them based on your preference.

---

### Regarding Converting vs Native Support

Thank you for pointing this out — I hadn't fully considered the implications of my initial approach.

After reflecting on your question, I realize that my current implementation has limitations:

- The digest changes after conversion
- Models cannot be pushed back in their original ModelPack format
- This breaks the complete push/pull/run cycle you mentioned

**Your suggestion of "pulling the format as-is without conversion"** makes a lot of sense for maintaining artifact integrity and full interoperability.

I've analyzed the codebase further and believe this can be achieved by:
- Storing the original ModelPack format (preserving the digest)
- Converting at **read time** instead of **pull time**
- This would require changes to only a few files, with no modifications needed for the scheduler or inference backends

---

### My Question

Would you prefer that I refactor the implementation to support native ModelPack storage as described above? 

I want to make sure I'm aligned with the project's direction before proceeding. Please let me know your thoughts — I'm happy to adjust my approach based on your guidance.

Thank you again for your patience and valuable feedback!

thc1006:
@ericcurtin

**TL;DR:** Refactored to **read-time conversion** as discussed — ModelPack is now stored as-is, preserving the original digest.

---

### What Changed

| Before | After |
|--------|-------|
| Convert at pull time | Convert at read time |
| Digest changes | Digest preserved ✅ |
| artifact.go wrapper (+666 lines) | Deleted (-666 lines) ✅ |

### Implementation

- `IsModelPackConfig()` — detects format via JSON structure parsing
- `partial.ConfigFile()` — auto-converts when reading config
- `MapLayerMediaType()` — maps layer types when querying layers
- Stored artifact remains unchanged (push/pull/run cycle works)

---

This addresses the concerns you raised. Ready for review when you have a chance. Thank you!

ericcurtin:
Need to remove non-English comments

ericcurtin:
We should treat this as a new version of the model format in DMR codebase and if it eventually matures, start to push/pull in that format. If the implementation matures, we could eventually switch to this as the primary format, given no negative consequences of the change (for example the existing DMR format treats layers as files to avoid duplication on disk, haven't checked if ModelPack does similar).

We should not convert between the two types.

No conversions, push and pull as the existing format. Push and pull as the CNCF format should be supported.

thc1006:
@ericcurtin

**TL;DR:** I'd like to confirm the scope of "no conversions" before proceeding further.

---

Thank you for your patience and continued guidance on this PR.

I've removed the non-English comments and rebased on the latest main. However, before making additional changes, I want to make sure I correctly understand your requirements.

You mentioned:
> "We should not convert between the two types"

Currently, the implementation:
- Stores ModelPack format as-is (original bytes, original digest preserved)
- Pushes back the original format unchanged
- At read-time, parses the ModelPack JSON into a common struct so the scheduler can understand it

My question is: does "no conversions" mean:

**(A)** Storage/transmission should not be modified (which is already the case), and read-time parsing into a common struct is acceptable?

or

**(B)** The scheduler and backends should natively understand both formats without any internal normalization?

If (B), I'm happy to implement a unified interface approach, though it would require more extensive changes.

I want to avoid going in the wrong direction, so your clarification would be very helpful. Thank you for your time.

ericcurtin:
(B) thanks for working on this, there is now a conflict also

PR #525, opened 2025-12-18
chore: remove unused memory estimation code and gpuinfo package

doringeman:
Cleanup after #504 which made this code unused. Remove GetRequiredMemoryForModel from Backend interface and all implementations, along with supporting functions in llamacpp and the entire gpuinfo package.


ilopezluna (approved):
oh there was much more! thanks :D

PR #530, opened 2025-12-22
Add openaiurl flag to list and run commands

ericcurtin:
The openaiurl flag enables users to connect to external
OpenAI-compatible API endpoints for both listing models and running chat
interactions. When this flag is specified, the commands will bypass the
local model runner and communicate directly with the provided endpoint.
The list command now supports filtering models from external endpoints,
and the run command supports both single prompt mode and interactive
mode with external endpoints.

### DIFF
--- a/cmd/cli/commands/list.go
+++ b/cmd/cli/commands/list.go
@@ -21,6 +21,7 @@ import (
 
 func newListCmd() *cobra.Command {
 	var jsonFormat, openai, quiet bool
+	var openaiURL string
 	c := &cobra.Command{
 		Use:     "list [OPTIONS] [MODEL]",
 		Aliases: []string{"ls"},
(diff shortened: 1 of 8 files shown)


ericcurtin:
This is like a reduced version of:

https://github.com/docker/model-runner/pull/251

I find this useful for testing various backends, llamacpp, vllm, etc.

cmd/cli/commands/list.go
```diff
@@ -239,3 +281,31 @@ func appendRow(table *tablewriter.Table, tag string, model dmrm.Model) {
...
+// prettyPrintOpenAIModels formats OpenAI model list in table format with only MODEL NAME populated
+func prettyPrintOpenAIModels(models dmrm.OpenAIModelList) string {
+	// Sort models by ID
+	sort.Slice(models.Data, func(i, j int) bool {
+		return strings.ToLower(models.Data[i].ID) < strings.ToLower(models.Data[j].ID)
+	})
+
+	var buf bytes.Buffer
+	table := newTable(&buf)
+	table.Header([]string{"MODEL NAME", "PARAMETERS", "QUANTIZATION", "ARCHITECTURE", "MODEL ID", "CREATED", "CONTEXT", "SIZE"})
```
doringeman:
```suggestion
	table.Header([]string{"MODEL NAME", "CREATED"})
```

cmd/cli/commands/list.go
```diff
@@ -239,3 +281,31 @@ func appendRow(table *tablewriter.Table, tag string, model dmrm.Model) {
...
+	for _, model := range models.Data {
+		table.Append([]string{
+			model.ID,
+			"",
+			"",
+			"",
+			"",
+			"",
+			"",
+			"",
```
doringeman:
```suggestion
			units.HumanDuration(time.Since(time.Unix(model.Created, 0))) + " ago",
```

doringeman (approved):
LGTM! Why empty fields in the table though?
```
$ docker model ls --openaiurl http://localhost:8080/v1
MODEL NAME          CREATED
ai/llama3.2:latest  9 months ago
```

ericcurtin:
I did implement this as you say here at one point. Then I changed my mind, there is a parsing advantage to keeping the columns consistent also. Lets go with what you suggested here.

PR #535, opened 2025-12-24
Add ability to pull vllm-compatible hf models

ericcurtin:
This commit introduces native HuggingFace model support by adding a new HuggingFace client implementation that can download safetensors files directly from HuggingFace Hub repositories. The changes include:

A new HuggingFace client with authentication, file listing, and download capabilities. The client handles LFS files, error responses, and rate limiting appropriately.

A downloader component that manages parallel file downloads with progress reporting and temporary file storage. It includes progress tracking and concurrent download limiting.

Model building functionality that downloads files from HuggingFace repositories and constructs OCI model artifacts using the existing builder framework.

Repository utilities for file classification, filtering, and size calculations to identify safetensors and config files needed for model construction.

Integration with the existing pull mechanism to detect HuggingFace references and attempt native pulling when no OCI manifest is found. This preserves existing OCI functionality while adding fallback support for raw HuggingFace repositories.

### DIFF
--- a/pkg/distribution/distribution/client.go
+++ b/pkg/distribution/distribution/client.go
@@ -6,9 +6,11 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"os"
 	"slices"
 	"strings"
 
+	"github.com/docker/model-runner/pkg/distribution/huggingface"
 	"github.com/docker/model-runner/pkg/distribution/internal/progress"
 	"github.com/docker/model-runner/pkg/distribution/internal/store"
 	"github.com/docker/model-runner/pkg/distribution/registry"
@@ -162,9 +164,10 @@ func (c *Client) normalizeModelName(model string) string {
 		return model
 	}
 
-	// Normalize HuggingFace model names (lowercase path)
+	// Normalize HuggingFace model names
 	if strings.HasPrefix(model, "hf.co/") {
 		// Replace hf.co with huggingface.co to avoid losing the Authorization header on redirect.
+		// Lowercase for OCI compatibility (repository names must be lowercase)
 		model = "huggingface.co" + strings.ToLower(strings.TrimPrefix(model, "hf.co"))
 	}
 
@@ -261,21 +264,31 @@ func (c *Client) resolveID(id string) string {
 
 // PullModel pulls a model from a registry and returns the local file path
 func (c *Client) PullModel(ctx context.Context, reference string, progressWriter io.Writer, bearerToken ...string) error {
+	// Store original reference before normalization (needed for case-sensitive HuggingFace API)
+	originalReference := reference
 	// Normalize the model reference
 	reference = c.normalizeModelName(reference)
 	c.log.Infoln("Starting model pull:", utils.SanitizeForLog(reference))
 
 	// Use the client's registry, or create a temporary one if bearer token is provided
 	registryClient := c.registry
+	var token string
 	if len(bearerToken) > 0 && bearerToken[0] != "" {
+		token = bearerToken[0]
 		// Create a temporary registry client with bearer token authentication
-		auth := &authn.Bearer{Token: bearerToken[0]}
+		auth := &authn.Bearer{Token: token}
 		registryClient = registry.FromClient(c.registry, registry.WithAuth(auth))
 	}
 
 	// First, fetch the remote model to get the manifest
 	remoteModel, err := registryClient.Model(ctx, reference)
 	if err != nil {
+		// Check if this is a HuggingFace reference and the error indicates no OCI manifest
+		if isHuggingFaceReference(reference) && isNotOCIError(err) {
+			c.log.Infoln("No OCI manifest found, attempting native HuggingFace pull")
+			// Pass original reference to preserve case-sensitivity for HuggingFace API
+			return c.pullNativeHuggingFace(ctx, originalReference, progressWriter, token)
+		}
 		return fmt.Errorf("reading model from registry: %w", err)
 	}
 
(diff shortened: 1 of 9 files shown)


pkg/distribution/huggingface/model.go
```diff
@@ -0,0 +1,159 @@
...
+
+	// Create builder from safetensors files
+	b, err := builder.FromSafetensors(safetensorsPaths)
+	if err != nil {
+		return nil, fmt.Errorf("create builder: %w", err)
+	}
+
+	// Create config archive if we have config files
+	if len(configFiles) > 0 {
+		configArchive, err := createConfigArchive(localPaths, configFiles, tempDir)
```
ilopezluna, line 89: `configArchive` should be removed once it is no longer needed. This is because we are using `CreateConfigArchiveInDir`, and its description indicates that it is the caller’s responsibility to remove the temporary file in order to prevent unnecessary storage usage.

pkg/distribution/huggingface/repository.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// Filename returns the base filename without directory path
+func (f *RepoFile) Filename() string {
+	return path.Base(f.Path)
+}
+
+// configExtensions defines file extensions treated as config files
+// This matches the existing packaging/safetensors.go logic
+var configExtensions = []string{".md", ".txt", ".json", ".vocab", ".jinja"}
```
ilopezluna: We should extract configExtensions, as it is currently duplicated here and in `safetensors.go`.

pkg/distribution/huggingface/repository.go
```diff
@@ -0,0 +1,119 @@
...
+			return FileTypeConfig
+		}
+	}
+
+	return FileTypeUnknown
+}
+
+// FilterModelFiles filters repository files to only include files needed for model-runner
+// Returns safetensors files and config files separately
+func FilterModelFiles(files []RepoFile) (safetensors []RepoFile, configs []RepoFile) {
```
ilopezluna, line 79: Oh I like this approach of having this `FilterModelFiles` and `ClassifyFile` to determine wether a file must be included in the OCI Artifact 👏 . I think we should unify this logic to be used also in `PackageFromDirectory`. But it can be done in a follow up PR. Let me know if you prefer address it here or in a follow up PR (I can work on the follow up PR)

pkg/distribution/huggingface/repository.go
```diff
@@ -0,0 +1,119 @@
...
+	// FileTypeUnknown is an unrecognized file type
+	FileTypeUnknown FileType = iota
+	// FileTypeSafetensors is a safetensors model weight file
+	FileTypeSafetensors
+	// FileTypeConfig is a configuration file (json, txt, etc.)
+	FileTypeConfig
+)
+
+// ClassifyFile determines the file type based on filename
+func ClassifyFile(filename string) FileType {
```
ilopezluna: No need to expose this function

pkg/distribution/huggingface/repository.go
```diff
@@ -0,0 +1,119 @@
...
+func TotalSize(files []RepoFile) int64 {
+	var total int64
+	for _, f := range files {
+		total += f.ActualSize()
+	}
+	return total
+}
+
+// IsSafetensorsModel checks if the files contain at least one safetensors file
+func IsSafetensorsModel(files []RepoFile) bool {
```
ilopezluna: No need to expose this function

pkg/distribution/huggingface/model.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("download files: %w", err)
+	}
+
+	// Step 4: Build the model artifact
+	if progressWriter != nil {
+		_ = progress.WriteProgress(progressWriter, "Building model artifact...", 0, 0, 0, "")
+	}
+
+	model, err := buildModelFromFiles(result.LocalPaths, safetensorsFiles, configFiles, tempDir)
```
ilopezluna, line 60: At this point, once all files have been downloaded I think we could (should) reuse the existing logic to package the model. This way we would have unified the process of packaging Safetensors models, otherwise it's duplicated.
We can also do it in a follow up PR as this PR is already useful for users

pkg/distribution/distribution/client.go
```diff
@@ -162,10 +164,11 @@ func (c *Client) normalizeModelName(model string) string {
 		return model
 	}
 
-	// Normalize HuggingFace model names (lowercase path)
+	// Normalize HuggingFace model names
 	if strings.HasPrefix(model, "hf.co/") {
 		// Replace hf.co with huggingface.co to avoid losing the Authorization header on redirect.
-		model = "huggingface.co" + strings.ToLower(strings.TrimPrefix(model, "hf.co"))
+		// Note: We preserve case since HuggingFace's native API is case-sensitive
+		model = "huggingface.co" + strings.TrimPrefix(model, "hf.co")
```
ilopezluna: Removing the `strings.ToLower` will break current implementation of pulling GGUF models from HF, because we will fail to parse the OCI reference as uppercase are not allowed in repository names. This error will be wrongly interpreted as `// Check if this is a HuggingFace reference and the error indicates no OCI manifest`and it will try to pull a native HuggingFace model. And it will fail because will not find safetensros models:
```
MODEL_RUNNER_HOST=http://localhost:13434 docker model pull hf.co/unsloth/Qwen3-0.6B-GGUF
Failed to pull model: Error: no safetensors files found in repository unsloth/Qwen3-0.6B-GGUF
```
The problem is this function:
```
// isNotOCIError checks if the error indicates the model is not OCI-formatted
// This happens when the HuggingFace repository doesn't have an OCI manifest
func isNotOCIError(err error) bool {
```
The assumption of "This happens when the HuggingFace repository doesn't have an OCI manifest" is wrong, because when we get a `registry.ErrInvalidReference` the repo could have an OCI manifest (in case of GGUF) but is failing because the reference is not OCI compliant

ilopezluna (commented):
This looks really promising! I think we should consider using this approach for GGUF files as well, because it gives us much more flexibility. For example, right now if we update the manifest (e.g. by adding a new config field), we depend on HF to implement it.

With this approach, we won’t depend on anyone else.

I’ve added comments to reduce duplication, but this can be addressed in follow-up PRs. However, I do see a blocking issue, because these changes break the current HF integration for pulling GGUF models. I can help fix the issue, let me know if you want to tackle it yourself 🙏

pkg/distribution/huggingface/model.go
ericcurtin: I've clarified the comment to explain the lifecycle

pkg/distribution/huggingface/repository.go
ericcurtin: Done. I've exported ConfigExtensions and SpecialConfigFiles from the packaging package and updated huggingface/repository.go to import them instead of defining duplicates.

pkg/distribution/huggingface/repository.go
ericcurtin: Agreed, happy to tackle this in a follow-up PR. The shared ConfigExtensions and SpecialConfigFiles constants are now in place which should make that unification straightforward.

pkg/distribution/huggingface/repository.go
ericcurtin: Done

pkg/distribution/huggingface/repository.go
ericcurtin: Done

pkg/distribution/huggingface/model.go
ericcurtin: Agreed, this would reduce duplication. Happy to address this in a follow-up PR as suggested.

pkg/distribution/distribution/client.go
ericcurtin: I altered

PR #536, opened 2025-12-27
Replace go-containerregistry with containerd/moby

ericcurtin:
This commit removes the vendored go-containerregistry package and replaces it with containerd and moby packages for OCI registry operations.

### DIFF
--- a/pkg/distribution/oci/remote/remote.go
+++ b/pkg/distribution/oci/remote/remote.go
@@ -346,7 +346,7 @@ func (f *manifestFetcher) Fetch(ctx context.Context, desc v1.Descriptor) (io.Rea
 
 	// For HuggingFace, try /manifests/ first for any JSON-like content
 	// since they don't serve manifests via /blobs/ at all
-	shouldUseManifestEndpoint := isHF && (desc.MediaType == "application/json" || strings.Contains(desc.MediaType, "+json"))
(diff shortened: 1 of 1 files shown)


pkg/distribution/oci/remote/remote.go
```diff
@@ -0,0 +1,782 @@
...
+	manifest    *oci.Manifest
+	rawManifest []byte
+	store       content.Store
+	ctx         context.Context
+	mu          sync.Mutex
+}
+
+// createResolver creates a docker resolver with the given options.
+func createResolver(o *options, ref reference.Reference) remotes.Resolver {
+	authorizer := docker.NewDockerAuthorizer(
```
ilopezluna, line 419: We receive the user agent in o, but it isn’t used. The user agent is required to pull models from Hugging Face; otherwise, they won’t return an OCI artifact.

I believe we also use the user agent for metrics, so it’s important to preserve it when interacting with the OCI registry.

cc @doringeman please keep me honest 🙏

ilopezluna (commented):
I've been testing basic functionality (pulling, packaging, running models, df, ls, etc) and they all work except pulling from HF.
What are the *.test files included? How I can use/test them?

ericcurtin:
*.test files removed

Issue #539, opened 2026-01-06
Support for OpenAI Responses API


mikesir87:
## Summary

Add support for the OpenAI Responses API - https://platform.openai.com/docs/api-reference/responses

## Reasoning

The Responses API is starting to become the default endpoint used by libraries. For example, the [`@ai-sdk/openai` library's docs](https://ai-sdk.dev/providers/ai-sdk-providers/openai) state:

> Since AI SDK 5, the OpenAI responses API is called by default (unless you specify e.g. 'openai.chat')

While they have fallbacks, higher-level abstractions may not provide the level of configuration needed to swap the endpoint being used.

PR #540, opened 2026-01-06
Add OpenAI Responses API compatibility layer

ericcurtin:
The OpenAI Responses API provides a stateful interface combining chat
completions with conversation management and tool use. This change
introduces a new API endpoint at /responses that transforms requests to
the standard chat completion format and handles both streaming and
non-streaming responses. Includes in-memory storage with TTL for
response state management and comprehensive test coverage.

### DIFF
--- a/main.go
+++ b/main.go
@@ -23,6 +23,7 @@ import (
 	"github.com/docker/model-runner/pkg/metrics"
 	"github.com/docker/model-runner/pkg/middleware"
 	"github.com/docker/model-runner/pkg/ollama"
+	"github.com/docker/model-runner/pkg/responses"
 	"github.com/docker/model-runner/pkg/routing"
 	"github.com/sirupsen/logrus"
 )
@@ -165,6 +166,16 @@ func main() {
 	router.Handle(inference.ModelsPrefix, modelHandler)
 	router.Handle(inference.ModelsPrefix+"/", modelHandler)
 	router.Handle(inference.InferencePrefix+"/", schedulerHTTP)
+	// Add OpenAI Responses API compatibility layer
+	responsesHandler := responses.NewHTTPHandler(log, schedulerHTTP, nil)
+	router.Handle(responses.APIPrefix+"/", responsesHandler)
+	router.Handle(responses.APIPrefix, responsesHandler) // Also register for exact match without trailing slash
+	router.Handle("/v1"+responses.APIPrefix+"/", responsesHandler)
+	router.Handle("/v1"+responses.APIPrefix, responsesHandler)
+	// Also register Responses API under inference prefix to support all inference engines
+	router.Handle(inference.InferencePrefix+responses.APIPrefix+"/", responsesHandler)
+	router.Handle(inference.InferencePrefix+responses.APIPrefix, responsesHandler)
+
 	// Add path aliases: /v1 -> /engines/v1, /rerank -> /engines/rerank, /score -> /engines/score.
 	aliasHandler := &middleware.AliasHandler{Handler: schedulerHTTP}
 	router.Handle("/v1/", aliasHandler)
--- a/pkg/inference/models/adapter.go
+++ b/pkg/inference/models/adapter.go
@@ -1,6 +1,7 @@
 package models
 
 import (
+	"encoding/json"
 	"fmt"
 
 	"github.com/docker/model-runner/pkg/distribution/types"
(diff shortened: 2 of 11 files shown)


ericcurtin:
https://github.com/docker/model-runner/issues/539

pkg/inference/models/adapter.go
```diff
@@ -27,12 +28,23 @@ func ToModel(m types.Model) (*Model, error) {
...
-	}, nil
+	}
+
+	// Marshal the config to populate RawConfig
+	if cfg != nil {
+		configData, err := json.Marshal(cfg)
+		if err != nil {
+			return nil, fmt.Errorf("marshal config: %w", err)
+		}
+		model.RawConfig = configData
```
ilopezluna, line 44: This should not be needed, the Responses API handler can work with the current Model struct without needing RawConfig. Unless I'm missing something no changes are needed in `adapter.go` and `api.go`

pkg/inference/models/api.go
```diff
@@ -112,7 +112,35 @@ type Model struct {
 	Created int64 `json:"created"`
 	// Config describes the model. Can be either Docker format (*types.Config)
 	// or ModelPack format (*modelpack.Model).
-	Config types.ModelConfig `json:"config"`
+	Config types.ModelConfig `json:"-"`
```
ilopezluna, line 115: As mentioned before I think these changes are not needed because unmarshalling has been fixed [in here](https://github.com/docker/model-runner/commit/c6e9171dc838a2c8085c11476690740be6a37448). Am I missing something?

pkg/inference/models/api.go
```diff
@@ -112,7 +112,35 @@ type Model struct {
...
+	// RawConfig is used for JSON marshaling/unmarshaling
+	RawConfig json.RawMessage `json:"config"`
+}
+
+// MarshalJSON implements custom marshaling for Model
+func (m Model) MarshalJSON() ([]byte, error) {
```
ilopezluna, line 121: `TestModelJSONRoundTrip` already verifies correct behavior of marshalling, so this method is not needed.

main.go
```diff
@@ -165,6 +166,13 @@ func main() {
 	router.Handle(inference.ModelsPrefix, modelHandler)
 	router.Handle(inference.ModelsPrefix+"/", modelHandler)
 	router.Handle(inference.InferencePrefix+"/", schedulerHTTP)
+	// Add OpenAI Responses API compatibility layer
+	responsesHandler := responses.NewHTTPHandler(log, schedulerHTTP, nil)
+	router.Handle(responses.APIPrefix+"/", responsesHandler)
```
ilopezluna, line 171: I've noticed you are not including the `inference.InferencePrefix` is this on purpose? Do we only want to support Responses API for one specific inference engine? Or do we plan to use another mechanism to choose which inference engine to use? (maybe based on the model)

pkg/responses/api.go
```diff
@@ -0,0 +1,410 @@
...
+// Response represents a complete response from the API.
+type Response struct {
+	// ID is the unique identifier for this response.
+	ID string `json:"id"`
+
+	// Object is always "response".
+	Object string `json:"object"`
+
+	// CreatedAt is the Unix timestamp when the response was created.
+	CreatedAt float64 `json:"created_at"`
```
ilopezluna, line 116: They also include `completed_at` in the response: https://platform.openai.com/docs/api-reference/responses/object

pkg/responses/api.go
```diff
@@ -0,0 +1,410 @@
...
+
+	// Metadata is user-defined metadata.
+	Metadata map[string]string `json:"metadata"`
+
+	// ParallelToolCalls indicates if parallel tool calls were enabled.
+	ParallelToolCalls *bool `json:"parallel_tool_calls"`
+
+	// Temperature used for generation.
+	Temperature *float64 `json:"temperature"`
+
```
ilopezluna, line 150: They also include `text` field https://platform.openai.com/docs/api-reference/responses/object#responses-object-text

pkg/responses/transform.go
```diff
@@ -0,0 +1,449 @@
...
+	Temperature       *float64      `json:"temperature,omitempty"`
+	TopP              *float64      `json:"top_p,omitempty"`
+	MaxTokens         *int          `json:"max_tokens,omitempty"`
+	Stream            bool          `json:"stream,omitempty"`
+	User              string        `json:"user,omitempty"`
+	ParallelToolCalls *bool         `json:"parallel_tool_calls,omitempty"`
+}
+
+// ChatMessage represents a message in the chat completion format.
+type ChatMessage struct {
```
ilopezluna, line 24: Not directly related to this PR (because we already have this same type in other place), but I'm interested on your opinion. Should we include https://github.com/openai/openai-go and use their types?

ericcurtin: Maybe, we'd have to investigate I guess. I do want to start being more careful about dependencies, as an example, I don't think we should have included this one from the start:

https://github.com/google/go-containerregistry

moby/containerd from the docker ecosystem had everything we ever needed and go-containerregistry isn't well maintained. Dependancies tend to increase our attack surface for CVEs also.

ericcurtin:
@ilopezluna I addressed some of the review comments in the last push. Merging.

Issue #472, opened 2025-12-03
Unfork llama.cpp


ericcurtin:
We should be able to use the code, cmake files, etc. from llama.cpp as is:

https://github.com/docker/model-runner/pull/471

doringeman:
We have to work around the patch we have which is for checking the readiness: https://github.com/docker/model-runner/blob/ac90dd36af98f3f47b70d5ef8545b6dd4e1500ae/pkg/inference/scheduling/runner.go#L213

ericcurtin:
@doringeman you seem like you are in the know here, I can do this to vanilla llama-server (which we should be using) no problem, what is difference about the fork that we cannot do similar?

```
$ curl -k http://127.0.0.1:8080/v1/models
{"models":[{"name":"unsloth/Qwen3-Coder-30B-A3B-Instruct-GGUF:Q4_K_M","model":"unsloth/Qwen3-Coder-30B-A3B-Instruct-GGUF:Q4_K_M","modified_at":"","size":"","digest":"","type":"model","description":"","tags":[""],"capabilities":["completion"],"parameters":"","details":{"parent_model":"","format":"gguf","family":"","families":[""],"parameter_size":"","quantization_level":""}}],"object":"list","data":[{"id":"unsloth/Qwen3-Coder-30B-A3B-Instruct-GGUF:Q4_K_M","object":"model","created":1764770422,"owned_by":"llamacpp","meta":{"vocab_type":2,"n_vocab":151936,"n_ctx_train":262144,"n_embd":2048,"n_params":30532122624,"size":18550716416}}]}
```

PR #541, opened 2026-01-06
Remove llama.cpp server fork and build upstream

ericcurtin:
This change removes the custom server fork at llamacpp/native/src/server
and instead builds the upstream llama.cpp server directly. The CMake
build is updated to copy the upstream llama-server binary to the
expected com.docker.llama-server name. The README is updated to reflect
that no custom patches are applied.

### DIFF
--- a/cmd/cli/commands/list.go
+++ b/cmd/cli/commands/list.go
@@ -261,7 +261,7 @@ func appendRow(table *tablewriter.Table, tag string, model dmrm.Model) {
 	contextSize := ""
 	if model.Config.GetContextSize() != nil {
 		contextSize = fmt.Sprintf("%d", *model.Config.GetContextSize())
-	} else if dockerConfig, ok := model.Config.ModelConfig.(*types.Config); ok && dockerConfig.GGUF != nil {
+	} else if dockerConfig, ok := model.Config.(*types.Config); ok && dockerConfig.GGUF != nil {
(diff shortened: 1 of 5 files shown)


cmd/cli/commands/list.go
```diff
@@ -261,7 +261,7 @@ func appendRow(table *tablewriter.Table, tag string, model dmrm.Model) {
 	contextSize := ""
 	if model.Config.GetContextSize() != nil {
 		contextSize = fmt.Sprintf("%d", *model.Config.GetContextSize())
-	} else if dockerConfig, ok := model.Config.(*types.Config); ok && dockerConfig.GGUF != nil {
+	} else if dockerConfig, ok := model.Config.ModelConfig.(*types.Config); ok && dockerConfig.GGUF != nil {
```
ilopezluna: addressed [in here](https://github.com/docker/model-runner/pull/542) the `ModelConfig` wrapper is not needed. 
Changes in `list.go`,`list_test.go`, `api.go`, `api.go.orig`,`adapter.go` and `api_test.go` are not needed anymore

ericcurtin:
@ilopezluna addressed, thanks for the review

doringeman:
It's still lacking some changes for this to work alright.
See https://github.com/docker/model-runner/issues/472#issuecomment-3606500179.
Also, the CMakeLists was changing the RPATH.
I've opened https://github.com/docker/model-runner/pull/546 to fix this.

PR #549, opened 2026-01-09
Add search command for Docker Hub and HuggingFace

ericcurtin:
Adds a new `docker model search` command that allows users to search for AI models from both Docker Hub and HuggingFace. The command supports filtering by search terms, limiting results, selecting specific sources (Docker Hub, HuggingFace, or both), and outputting results in either table or JSON format. Implements concurrent search across multiple sources with proper error handling and rate limiting support.


cmd/cli/search/dockerhub.go
```diff
@@ -0,0 +1,203 @@
...
+	for len(results) < limit {
+		var fullURL string
+		if nextURL != "" {
+			fullURL = nextURL
+		} else {
+			// Build the URL for listing repositories in the ai/ namespace
+			apiURL := fmt.Sprintf("%s/v2/repositories/%s/", c.baseURL, dockerHubAIOrg)
+			params := url.Values{}
+			params.Set("page_size", fmt.Sprintf("%d", pageSize))
+			params.Set("ordering", "-pull_count") // Sort by popularity
```
ilopezluna:
```suggestion
			params.Set("ordering", "pull_count") // Sort by popularity
```
Seems like the API is wrong https://docs.docker.com/reference/api/hub/latest/#tag/repositories/operation/listNamespaceRepositories

ilopezluna (approved):

PR #563, opened 2026-01-13
chore(ci): improve release workflow

doringeman:
- Add job summary table showing all release parameters at a glance
- Simplify tag formatting logic using a reusable bash function (42 → 18 lines)
- Make releaseTag a required input to prevent accidental pushes to default "test" tag

Test: https://github.com/docker/model-runner/actions/runs/20959387670.

Screenshot 2026-01-13 at 15 59 16

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,9 +11,8 @@ on:
         default: false
       releaseTag:
         description: "Release tag"
-        required: false
+        required: true
         type: string
-        default: "test"
       llamaServerVersion:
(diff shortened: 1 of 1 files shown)


ericcurtin:
How about?

scripts/release-parameters.sh
scripts/format-tags.sh

rather than embedded bash? We can run linters like shellcheck, we get syntax highlighting and can run more easily locally (although I understand not all of these things are suitable for running locally, the other two benefits still apply).

doringeman:
I agree, will do!

PR #569, opened 2026-01-14
refactor(distribution): use shared registry client instead of duplicating options

doringeman:
Replace WithTransport, WithUserAgent, WithPlainHTTP, and WithRegistryAuth options with a single WithRegistryClient option. This eliminates the duplicate registry client creation - callers now create one registry client and pass it to both the distribution client and use it directly.


ilopezluna (approved):
This looks much better! thanks for working on this 🙏

PR #544, opened 2026-01-07
Add diffusers backend support for image generation

ericcurtin:
This commit introduces a new diffusers backend that enables image
generation capabilities in the model runner. The backend integrates with
Hugging Face diffusers library to provide OpenAI-compatible image
generation API endpoints.

The implementation includes a new Python server component that runs as a
subprocess and handles image generation requests through a Unix socket
interface. The backend supports various diffusers models and provides
configuration options for device selection, precision, and memory
optimization features.

Additionally, the Dockerfile now includes a diffusers variant with the
necessary Python dependencies pre-installed. The model distribution
logic has been updated to properly handle diffusers model formats with
their directory structure preserved in the model bundles.

Signed-off-by: Eric Curtin <eric.curtin@docker.com>

PR #572, opened 2026-01-15
Add stable diffusion backend

ilopezluna:
Based on https://github.com/docker/model-runner/pull/544 
I started a new branch because the previous PR relied on `gocontainerregistry`. Given that, it was easier to start from scratch and use the previous PR as a reference rather than updating the old one.

In this PR, we also add support for packaging Diffusers models. For now, we only support the DDUF format (similar to GGUF: a single file containing everything), as it was the easiest to implement. However, support for other formats can be added later without much difficulty.

The backend is currently enabled only on Linux, but it also works on Darwin. That said, it requires a Python runtime, so we should first design a solution that works well in Docker Desktop.

I have packaged and pushed `ignaciolopezluna020/stable-diffusion:latest` and `ignaciolopezluna020/stable-diffusion:Q4`.

You can try by:

Building the image: 
`docker build --target final-diffusers -t model-runner:diffusers .`

Then running it:
`docker run -it --rm --gpus all -p 13434:13434 -e MODEL_RUNNER_PORT=13434 model-runner:diffusers`

Pull model:
`MODEL_RUNNER_HOST=http://localhost:13434 docker model pull ignaciolopezluna020/stable-diffusion:q4`

And run it:
```
curl -s -X POST http://localhost:13434/engines/diffusers/v1/images/generations \
  -d '{
    "model": "ignaciolopezluna020/stable-diffusion:Q4",
    "prompt": "A cat sitting on a couch",
    "size": "512x512"
  }' | jq -r '.data[0].b64_json' | base64 -d > cat.png
```

Note: the image is being built from llama.cpp one, it does not include vllm

### DIFF
--- a/scripts/docker-run.sh
+++ b/scripts/docker-run.sh
@@ -5,7 +5,9 @@ add_accelerators() {
   if [[ "${DOCKER_IMAGE-}" == *"-cuda" ]] || \
      [[ "${DOCKER_IMAGE-}" == *"-diffusers" ]] || \
      [[ "${DOCKER_IMAGE-}" == *"-sglang" ]]; then
-    args+=("--gpus" "all" "--runtime=nvidia")
+      if docker info -f '{{range $k, $v := .Runtimes}}{{$k}}{{"\n"}}{{end}}' 2>/dev/null | grep -qx "nvidia"; then
+        args+=("--gpus" "all" "--runtime=nvidia")
+      fi
   fi
 
   # Add GPU/accelerator devices if present


Dockerfile
```diff
@@ -75,7 +75,7 @@ ENV MODEL_RUNNER_PORT=12434
 ENV LLAMA_SERVER_PATH=/app/bin
 ENV HOME=/home/modelrunner
 ENV MODELS_PATH=/models
```
ilopezluna, line 77: I've removed `$LD_LIBRARY_PATH` because I didn't find any usage. Please let me know if I'm missing anything here 🙏

pkg/distribution/format/dduf.go
```diff
@@ -0,0 +1,77 @@
...
+		Diffusers: map[string]string{
+			"layout":    "dduf",
+			"dduf_file": ddufFile,
+		},
+	}, nil
+}
+
+// formatDDUFSize converts bytes to human-readable format matching Docker's style
+// Returns format like "256MB" (decimal units, no space, matching `docker images`)
+func formatDDUFSize(bytes int64) string {
```
doringeman: This duplicates the `formatSize` function from `safetensors.go`.
Can we use the same function for all 3 formats?

pkg/distribution/internal/bundle/unpack.go
```diff
@@ -88,9 +92,41 @@ func detectModelFormat(model types.Model) types.Format {
...
+		return fmt.Errorf("no DDUF files found")
+	}
+
+	modelDir := filepath.Join(bundle.dir, ModelSubdir)
+
+	// DDUF is a single-file format
+	ddufFilename := filepath.Base(ddufPaths[0])
+	// Ensure the filename has the .dduf extension for proper detection by diffusers server
+	if !strings.HasSuffix(strings.ToLower(ddufFilename), ".dduf") {
+		ddufFilename = ddufFilename + ".dduf"
```
doringeman, line 121: Is this really needed? Why don't we error out if doesn't have `.dduf`?

Dockerfile
```diff
@@ -144,6 +144,60 @@ RUN curl -LsSf https://astral.sh/uv/install.sh | sh \
     && ~/.local/bin/uv pip install --python /opt/sglang-env/bin/python "sglang==${SGLANG_VERSION}"
 
 RUN /opt/sglang-env/bin/python -c "import sglang; print(sglang.__version__)" > /opt/sglang-env/version
+
+# --- Diffusers variant ---
+FROM llamacpp AS diffusers
```
doringeman, line 149: You might be aware of this, but It'll be good to add `make docker-build-diffusers` and `make docker-run-diffusers` as it's done for the other variants.

pkg/distribution/internal/bundle/unpack.go
ilopezluna: This is during unpacking into a bundle. We use the SHA of the blob as a filename, so no `.dduf` extension. The diffusers fails if the file provided has no `.dduf` extension :(

doringeman:
You should add the new `image-generation` mode to the following files as well (search for completion/embedding/reranking):
- ./cmd/cli/commands/configure_flags.go
- ./cmd/cli/commands/compose_test.go
- ./pkg/inference/scheduling/loader.go

doringeman:
Currently it doesn't work on `arm64`. Do you plan to try to get it working in this PR or?

doringeman (approved):

PR #576, opened 2026-01-16
Add /set command and context length configuration

ericcurtin:
Add a /set command to the interactive mode to allow users to configure
parameters like num_ctx during runtime.

### DIFF
--- a/cmd/cli/commands/run.go
+++ b/cmd/cli/commands/run.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"os"
 	"os/signal"
+	"strconv"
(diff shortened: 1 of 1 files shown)


ilopezluna (commented):
`DMR_CONTEXT_LENGTH` has no effect unless the model is configured.
I’m mentioning this because it confused me while testing the PR. I initially set the environment variable and ran a model, expecting the context length to be taken from the env var. However, it still used the default context size, since the logic that checks this environment variable only runs during model configuration.
Just to confirm, is this the expected behaviour?

PR #581, opened 2026-01-19
(WIP) - Added layer wise reporting on push

MetsysEht:


pkg/distribution/huggingface/downloader.go
```diff
@@ -208,7 +208,7 @@ func (pr *progressReader) Read(p []byte) (n int, err error) {
 
 		// Report progress periodically (every 1MB or when complete)
 		if pr.progressWriter != nil && (pr.bytesRead-pr.lastReported >= progress.MinBytesForUpdate || pr.bytesRead == pr.fileSize) {
-			_ = progress.WriteProgress(pr.progressWriter, "", pr.totalImageSize, pr.fileSize, pr.bytesRead, pr.fileID)
+			_ = progress.WriteProgress(pr.progressWriter, "", pr.totalImageSize, pr.fileSize, pr.bytesRead, pr.fileID, "")
```
ilopezluna: Even when we default to `pull` if no mode is specified, I think its better to be explicit instead of using an empty string, this way its clear the intend

pkg/distribution/internal/progress/reporter.go
```diff
@@ -39,6 +40,7 @@ type Reporter struct {
 	format    progressF
 	layer     oci.Layer
 	imageSize uint64
+	mode      string
```
ilopezluna, line 43: oh this struct is duplicated (its the same as `ProgressMessage`). No need to address it in this PR we can do it in a follow up PR

pkg/distribution/huggingface/downloader.go
MetsysEht: Will the "pull" value be correct in this context? If not what value should I pass here then?

ilopezluna: `pull` value is correct in this context, this downloader is responsible of pulling models from Hugging Face

MetsysEht: I have updated the values

ilopezluna:
excellent progres @MetsysEht I think we are good to go! Could you please address the comments from the AI bot? the one from [sourcery-ai](https://github.com/apps/sourcery-ai)  seems worth to address 🙏

pkg/distribution/oci/remote/remote.go
```diff
@@ -921,6 +939,15 @@ func closeProgress(ch chan<- oci.Update) {
 	}
 }
 
+// closeReporter safely closes the progress reporter if not nil
+func closeReporter(pr *progress.Reporter) {
+	if pr != nil {
+		if waitErr := pr.Wait(); waitErr != nil {
```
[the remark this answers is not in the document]
MetsysEht: @ilopezluna 
1. Removed the total size calculation as it is no longer being used.
2. Yes, It is the desired behavior
3. Yes, Currently we are doing push 1 layer at a time, I was thinking of making it in go routines for parallel push, if that is needed.
4. I have followed the same close as in target.go, should I change it in both the places
https://github.com/docker/model-runner/blob/52088bdc7a0f6d74d9d65aa1772b28c473ec8e2a/pkg/distribution/tarball/target.go#L124-L126

Issue #445, opened 2025-11-24
Add TLS between model-cli and the Model Runner API when using TCP


ilopezluna:
When the Model Runner API is exposed over TCP (http://localhost:12434 by default), requests between model-cli and the API travel unencrypted. 
Add first-class support for TLS so the CLI can connect securely over https://…, aligning with Docker Engine’s existing TLS ergonomics. 
Today, Unix domain socket access exists (secure by OS permissions), but TCP is explicitly supported and documented for host access, which should be able to be secured as well.

ericcurtin:
+1 to this.

We may also want to explore some encryption techniques that are low on configuration and "just work".

TLS is the standard and we should add, but requires manual effort, issuing certificates, or PSK, etc.

The solution to this might just be put something like NGINX in front of DMR daemon and use the cli --tls options.

ericcurtin:
Maybe we just generate self-signed certs automatically and use TLS, it's better than plain-text HTTP

ilopezluna:
Addressed in https://github.com/docker/model-runner/pull/582

PR #582, opened 2026-01-19
Add TLS certificate generation, management utils

ericcurtin:
This introduces a new pkg/tls package that provides automatic TLS
certificate generation for the Model Runner API server. The feature
includes self-signed CA and server certificate creation with proper
validation and renewal checks. Additionally, TLS support is integrated
into the install-runner, reinstall-runner, and start-runner commands
with new flags for enabling TLS, specifying ports, and providing custom
certificate paths. The Model Runner context now supports TLS-enabled
connections with proper client configuration.

### DIFF
--- a/cmd/cli/pkg/standalone/containers.go
+++ b/cmd/cli/pkg/standalone/containers.go
@@ -274,8 +274,23 @@ func tryGetBindAscendMounts(printer StatusPrinter, debug bool) []mount.Mount {
 // This location is used by update-ca-certificates to add the cert to the system trust store.
 const proxyCertContainerPath = "/usr/local/share/ca-certificates/proxy-ca.crt"
 
+// TLSOptions holds TLS configuration for the controller container.
+type TLSOptions struct {
+	// Enabled indicates whether TLS is enabled.
+	Enabled bool
+	// Port is the TLS port (0 to use default).
+	Port uint16
+	// CertPath is the path to the TLS certificate file.
+	CertPath string
+	// KeyPath is the path to the TLS key file.
+	KeyPath string
+}
+
+// tlsCertContainerPath is the path where TLS certificates will be mounted in the container.
+const tlsCertContainerPath = "/etc/model-runner/certs"
+
 // CreateControllerContainer creates and starts a controller container.
-func CreateControllerContainer(ctx context.Context, dockerClient *client.Client, port uint16, host string, environment string, doNotTrack bool, gpu gpupkg.GPUSupport, backend string, modelStorageVolume string, printer StatusPrinter, engineKind types.ModelRunnerEngineKind, debug bool, vllmOnWSL bool, proxyCert string) error {
+func CreateControllerContainer(ctx context.Context, dockerClient *client.Client, port uint16, host string, environment string, doNotTrack bool, gpu gpupkg.GPUSupport, backend string, modelStorageVolume string, printer StatusPrinter, engineKind types.ModelRunnerEngineKind, debug bool, vllmOnWSL bool, proxyCert string, tlsOpts TLSOptions) error {
 	imageName := controllerImageName(gpu, backend)
 
 	// Set up the container configuration.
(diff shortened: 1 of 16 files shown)


pkg/tls/certs.go
```diff
@@ -0,0 +1,349 @@
...
+	if customPath != "" {
+		return customPath, nil
+	}
+
+	paths, err := DefaultCertPaths()
+	if err != nil {
+		return "", err
+	}
+	return paths.CACert, nil
+}
```
doringeman, line 349: Dead code.
```suggestion
```

pkg/tls/certs.go
```diff
@@ -0,0 +1,349 @@
...
+			Organization:       []string{"Docker Model Runner"},
+			OrganizationalUnit: []string{"Self-Signed CA"},
+			CommonName:         "Docker Model Runner CA",
+		},
+		NotBefore:             time.Now().Add(-1 * time.Hour), // Allow for clock skew
+		NotAfter:              time.Now().AddDate(0, 0, DefaultCAValidityDays),
+		IsCA:                  true,
+		KeyUsage:              x509.KeyUsageCertSign | x509.KeyUsageCRLSign | x509.KeyUsageDigitalSignature,
+		BasicConstraintsValid: true,
+		MaxPathLen:            1,
```
doringeman, line 192:
```suggestion
		MaxPathLen:            0,
		MaxPathLenZero:        true,
```
?
Why do we allow an intermediate CA?

pkg/tls/certs.go
```diff
@@ -0,0 +1,349 @@
...
+		Subject: pkix.Name{
+			Organization:       []string{"Docker Model Runner"},
+			OrganizationalUnit: []string{"Server"},
+			CommonName:         "localhost",
+		},
+		NotBefore:   time.Now().Add(-1 * time.Hour), // Allow for clock skew
+		NotAfter:    time.Now().AddDate(0, 0, DefaultCertValidityDays),
+		KeyUsage:    x509.KeyUsageDigitalSignature | x509.KeyUsageKeyEncipherment,
+		ExtKeyUsage: []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+		DNSNames:    []string{"localhost", "docker-model-runner"},
```
doringeman, line 236: I wonder if we'll need to add another one here for Offload.

cmd/cli/pkg/standalone/containers.go
```diff
@@ -296,12 +311,51 @@ func CreateControllerContainer(ctx context.Context, dockerClient *client.Client,
...
+		} else {
+			tlsPort = DefaultTLSPortMoby
+		}
+	}
+
+	// Add TLS environment variables if TLS is enabled
+	if tlsOpts.Enabled {
+		env = append(env, "MODEL_RUNNER_TLS_ENABLED=true")
+		env = append(env, "MODEL_RUNNER_TLS_PORT="+strconv.Itoa(int(tlsPort)))
+		if tlsOpts.CertPath != "" && tlsOpts.KeyPath != "" {
```
doringeman, line 328: It looks like both on the client side and on the server side, both certificate and key paths must be provided, or neither. But this is silently ignored and we even proceed with the auto-generated certs if only one is provided.

doringeman:
FWIW this fixes https://github.com/docker/model-runner/issues/445.

ericcurtin:
@doringeman I'm gonna merge this to unblock you if you want to look at this futher, feel free to make any suggested changes, even breaking ones, I'll approve the PR.

One thing that's important when we implement this, we have to retain the ability to do plain-text http also by default, because many clients/tools out there will assume http as if they were connection to a plain-text vllm, llama.cpp, Ollama, instance etc.

ericcurtin:
And yeah I did some autogenerating self-signed stuff here, that's not great, but it's better than no encryption. It's still encrypted, not authenticated. Even with this, it's more secure than most other inference engines which do zero and maybe doing zero is not an unwise choice, in a production environment, maybe you do just put nginx in front of this or something similar to do tls encryption and forget about it.

ericcurtin:
Somehow this broken the build on main, even though it passed here. A linter build

PR #584, opened 2026-01-20
Modernize tests

dgageot:
I tried to use a cagent based agent to modernize the Go unit tests as much as possible.


dgageot:
I'm working on fixing the linter issue. The rules I've added have the nasty side effect of changing the existing forbidigo config

doringeman (commented):
Thanks!

PR #586, opened 2026-01-20
replace ProgressMessage with oci.ProgressMessage and update related functions

ilopezluna:
This pull request refactors progress reporting across the codebase to standardize and centralize progress message types and handling. The main change is the migration of the `ProgressMessage` type and related constants to the `oci` package, replacing legacy and duplicated definitions.

### DIFF
--- a/pkg/distribution/oci/progress.go
+++ b/pkg/distribution/oci/progress.go
@@ -46,7 +46,7 @@ type ProgressLayer struct {
 // ProgressMessage represents a structured message for progress reporting
 type ProgressMessage struct {
 	Type    MessageType   `json:"type"`    // Message type: progress, success, warning, or error
-	Message string        `json:"message"` // Deprecated: the message should be defined by clients based on Message.Total and Message.Layer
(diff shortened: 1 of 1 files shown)


ilopezluna:
@MetsysEht I'm trying to refactor the progress types a bit, let me know if you have suggestions 🙏

MetsysEht:
@ilopezluna LGTM, Can we remove the deprecated Message Field and corresponding functions in the ProgressMessage struct as well. It causes confusion in the flow. Take a look at sourcery comment as well

ilopezluna:
> remove the deprecated Message Field and corresponding functions in the ProgressMessage struct as well. It causes confusion in the flow.

@MetsysEht I was tempted to remove it but unfortunately it would require more changes as we are using it to report warnings and errors :(

MetsysEht:
@ilopezluna Then we should update the comment to reflect that. Something like Deprecated for success/failed, still in use for warnings and errors

ilopezluna:
@MetsysEht added in https://github.com/docker/model-runner/pull/586/commits/1ece44702256a14125d3693bd42a8e57b73d7f9c

PR #590, opened 2026-01-22
refactor: resolve Docker credentials on CLI side instead of copying config

### LINKED ISSUE #560: OCI registry auth fails with remote Docker daemons due to backend credential resolution
### Problem

OCI registry credentials are currently resolved in the backend. This differs from Docker Images, where the client resolves credentials and passes them to the daemon.

Backend-side credential resolution fails when using a remote Docker daemon (e.g. Docker Offload), because the remote host has no access to the local `config.json`. As a result, pulling/pushing to authenticated OCI registries does not work with remote daemons.

### Expected behavior

OCI registry credentials should be resolved on the client, consistent with Docker Images, to support remote Docker daemons.

cc: @doringeman

doringeman:
Instead of copying ~/.docker/config.json into the container, credentials are now resolved on the CLI side using the Docker credential helpers and sent with pull/push requests. This simplifies container setup and avoids exposing the full Docker config inside the container.

To test following the next steps, you'd need to either connect to Offload or use `_MODEL_RUNNER_TREAT_DESKTOP_AS_MOBY=1`.

```
make docker-build DOCKER_IMAGE=docker/model-runner:test-docker-cli-auth
```

```
MODEL_RUNNER_CONTROLLER_VERSION=test-docker-cli-auth MODEL_RUNNER_CONTROLLER_VARIANT=cpu docker model start-runner
```
```
$ docker model package --gguf /Users/dorin/.cache/huggingface/hub/models--HuggingFaceTB--SmolLM2-360M-Instruct-GGUF/snapshots/593b5a2e04c8f3e4ee880263f93e0bd2901ad47f/smollm2-360m-instruct-q8_0.gguf --push doringeman657/my-smollm2
Adding GGUF file from "/Users/dorin/.cache/huggingface/hub/models--HuggingFaceTB--SmolLM2-360M-Instruct-GGUF/snapshots/593b5a2e04c8f3e4ee880263f93e0bd2901ad47f/smollm2-360m-instruct-q8_0.gguf"
Pushing model to registry...
WARN[0001] reference for unknown type: application/vnd.docker.ai.gguf.v3
Uploaded: 368.50 MB
Model pushed successfully
```

Resolves #560.
I'd merge this once we get https://github.com/docker/model-runner/pull/582 merged as well.

### DIFF
--- a/cmd/cli/desktop/desktop.go
+++ b/cmd/cli/desktop/desktop.go
@@ -16,6 +16,9 @@ import (
 
 	"github.com/docker/model-runner/cmd/cli/pkg/standalone"
 	"github.com/docker/model-runner/pkg/distribution/distribution"
+	"github.com/docker/model-runner/pkg/distribution/oci/authn"
+	"github.com/docker/model-runner/pkg/distribution/oci/reference"
+	"github.com/docker/model-runner/pkg/distribution/oci/remote"
 	"github.com/docker/model-runner/pkg/inference"
 	dmrm "github.com/docker/model-runner/pkg/inference/models"
 	"github.com/docker/model-runner/pkg/inference/scheduling"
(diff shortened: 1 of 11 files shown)


cmd/cli/desktop/desktop.go
```diff
@@ -223,12 +297,27 @@ func (c *Client) withRetries(
...
+			body = bytes.NewReader(jsonData)
+		}
+
 		pushPath := inference.ModelsPrefix + "/" + model + "/push"
 		resp, err := c.doRequest(
 			http.MethodPost,
 			pushPath,
-			nil, // Assuming no body is needed for the push request
+			body,
 		)
```
ilopezluna, line 321: why did you chose to send credentials through the body instead of using an auth header?

doringeman: Not being for the Model Runner API itself, I chose to not them in an auth header. They're for a separate/downstream  operation, hence I considered it'd be better to put them in the body. Moreover, if we ever want to set up authentication for Model Runner, then we'd use the auth header.
Do you think it'd be more suitable to have these registry creds in the auth header?

> I wonder if we should enable TLS by default. What do you think?

It sounds good, at least for the push/pull operations. I'd need to take a closer look though. Do you know already whether we can do this cleanly?

ilopezluna: > > I wonder if we should enable TLS by default. What do you think?
> 
> It sounds good, at least for the push/pull operations. I'd need to take a closer look though. Do you know already whether we can do this cleanly?

I'll take a look on how to use TLS by default on a separated branch 👍

doringeman: Will that branch be on top of this one? I think this would ease testing.

ilopezluna: From our previous discussion: 
> Because tools like llama-server, vllm serve, ollama, etc. don't use TLS. A lot of clients assume http, so we need plain-text http on also for this reason

PR #594, opened 2026-01-23
Bump llama.cpp to b7815

krissetto:


krissetto:
I'll rebase on main for the linter issue.

Huh i didn't notice the error because i used docker to build (and on linux). Will TAL

krissetto:
If i understood correctly, I think we could just remove `DLLAMA_CURL` since only macOS local builds are sandboxed, but i changed them all to `DLLAMA_OPENSSL` for sake of consistency. 
 
 Let me know if we prefer to do otherwise and I'll change it

doringeman:
They also removed `LLAMA_CURL`: https://github.com/ggml-org/llama.cpp/commit/516a4ca9b5f2fa72c2a71f412929a67cf76a6213#diff-ed99c076a4accf9ff1d7ad65a79dc3802df6bc8c91c9879e12c766a1b414272bL171.

Issue #526, opened 2025-12-18
Host-side TCP setting blocks vLLM backend installation


krissetto:
I'm running Docker Desktop on Windows and wanted to try vLLM. 

Running `docker model install-runner --backend=vllm --gpu=cuda` results in an error if the user previously already enabled host-side TCP support. 

What i see is the following:

<img width="1115" height="723" alt="Image" src="https://github.com/user-attachments/assets/fae6737f-e810-490b-85bd-c6cca95277b7" />

This can cause some friction if the user doesn't remember about the setting in Docker Desktop, also because `docker ps` does not show the inference server container.

Disabling the host-side TCP setting in Desktop and re-running the `install-runner` command solves the issue and the backend installs successfully

ilopezluna:
Added clarification message:

```
Starting model runner container docker-model-runner...
Error: unable to initialize standalone model runner container: failed to start container docker-model-runner: Error response from daemon: ports are not available: exposing port TCP 127.0.0.1:12434 -> 127.0.0.1:0: listen tcp4 127.0.0.1:12434: bind: address already in use

The port may already be in use by Docker Desktop's model runner.
Try running: docker desktop disable model-runner
```

tony-topper:
I am having a similar experience, but it leaves things in a strange state.  Not sure if I should file a separate bug or reopen this one.

After running `docker model install-runner --backend=vllm --gpu=cuda`

`docker model version`

returns

> Docker Model Runner version v1.0.12
> Docker Engine Kind: Docker Engine
> 

vllm appears to work but llama.cpp doesn't as shown by running `docker model status`

> Docker Model Runner is running
> 
> Status:
> llama.cpp: running llama.cpp version: unknown
> mlx: not installed
> sglang: sglang package not installed
> vllm: running vllm version: 0.12.0
> diffusers: diffusers package not installed


If I start DRM via the Windows desktop app, it is running something different.

`docker model version`

> Docker Model Runner version v1.0.12
> Docker Engine Kind: Docker Desktop

`docker model status`

> Docker Model Runner is running
> 
> Status:
> vllm: not installed
> llama.cpp: running llama.cpp latest-cuda (sha256:660c61310a6e2e82778ff2e3cffa439ebf9d8e9779cb8ce20eda0413ecdde4d5) version: c55bce4

Notice the two different "Docker Engine Kind"

[...]

PR #596, opened 2026-01-23
fix(containers): improve port binding error message for clarity

ilopezluna:
Address https://github.com/docker/model-runner/issues/526 by providing the command to disable DD TCP:
```
Starting model runner container docker-model-runner...
Error: unable to initialize standalone model runner container: failed to start container docker-model-runner: Error response from daemon: ports are not available: exposing port TCP 127.0.0.1:12434 -> 127.0.0.1:0: listen tcp4 127.0.0.1:12434: bind: address already in use

The port may already be in use by Docker Desktop's model runner.
Try running: docker desktop disable model-runner
```

### DIFF
--- a/cmd/cli/pkg/standalone/containers.go
+++ b/cmd/cli/pkg/standalone/containers.go
@@ -566,7 +566,7 @@ func CreateControllerContainer(ctx context.Context, dockerClient *client.Client,
 			_ = dockerClient.ContainerRemove(ctx, resp.ID, container.RemoveOptions{Force: true})
 		}
 		if isPortBindingError(err) {
-			return fmt.Errorf("failed to start container %s: %w\n\nThe port may already be in use by Docker Desktop's model runner.\nTry running: docker desktop disable model-runner", controllerContainerName, err)
+			return fmt.Errorf("failed to start container %s: %w\n\nThe port may already be in use by Docker Desktop's Model Runner.\nTry running: docker desktop disable model-runner", controllerContainerName, err)
 		}
 		return fmt.Errorf("failed to start container %s: %w", controllerContainerName, err)
 	}


cmd/cli/pkg/standalone/containers.go
```diff
@@ -554,6 +565,9 @@ func CreateControllerContainer(ctx context.Context, dockerClient *client.Client,
 		if created {
 			_ = dockerClient.ContainerRemove(ctx, resp.ID, container.RemoveOptions{Force: true})
 		}
+		if isPortBindingError(err) {
+			return fmt.Errorf("failed to start container %s: %w\n\nThe port may already be in use by Docker Desktop's model runner.\nTry running: docker desktop disable model-runner", controllerContainerName, err)
```
doringeman:
```suggestion
			return fmt.Errorf("failed to start container %s: %w\n\nThe port may already be in use by Docker Desktop's Model Runner.\nTry running: docker desktop disable model-runner", controllerContainerName, err)
```

cmd/cli/pkg/standalone/containers.go
```diff
@@ -289,6 +289,17 @@ type TLSOptions struct {
...
 
+// isPortBindingError checks if the error indicates a port is already in use
+func isPortBindingError(err error) bool {
+	if err == nil {
+		return false
+	}
+	errStr := err.Error()
+	return strings.Contains(errStr, "ports are not available") &&
+		(strings.Contains(errStr, "address already in use") ||
+			strings.Contains(errStr, "Only one usage of each socket address"))
```
doringeman, line 300: How do you get this one?

cmd/cli/pkg/standalone/containers.go
```diff
@@ -289,6 +289,17 @@ type TLSOptions struct {
 // tlsCertContainerPath is the path where TLS certificates will be mounted in the container.
 const tlsCertContainerPath = "/etc/model-runner/certs"
 
+// isPortBindingError checks if the error indicates a port is already in use
+func isPortBindingError(err error) bool {
```
doringeman, line 293: Maybe it'd be nice to actually check whether it's another DMR instance running on that port, perhaps by checking http://localhost:12434/engines/status.

cmd/cli/pkg/standalone/containers.go
ilopezluna: You can force it by using `install-runner` when DD has TCP enabled (its the same that @krissetto got: https://github.com/docker/model-runner/issues/526)

cmd/cli/pkg/standalone/containers.go
ilopezluna: @doringeman Actually that would require to do this extra http request every single time. Because it does only affect to DD users that  install vllm on WSL I prefer to not add the additional check and just add the hint in the error message.

PR #601, opened 2026-01-27
Filter runtime keys

ilopezluna:
This pull request implements an allowlist for runtime flags for different inference backends to improve security and prevent the use of dangerous flags, especially those involving file paths. 
The validation logic now checks each flag against a backend-specific allowlist and still enforces path safety as a defense-in-depth measure.


ilopezluna:
Closing this, as it has been decided not to support runtime flags.

ilopezluna:
@sourcery-ai dismiss

PR #632, opened 2026-02-05
fix(launch): improve error message for missing arguments

doringeman:
Now:
```
$ docker model launch
'docker model launch' requires at least 1 argument(s).

Usage:  docker model launch APP [-- APP_ARGS...]

See 'docker model launch --help' for more information
```

Before:
```
$ docker model launch
requires at least 1 arg(s), only received 0
```

CC @denyszhak


ericcurtin:
You should blog about this by the way @denyszhak it's a cool new feature

denyszhak:
> You should blog about this by the way @denyszhak it's a cool new feature

@ericcurtin that's an interesting idea:) I might exercise that option!

PR #638, opened 2026-02-05
feat: add vLLM ROCm variant for AMD GPUs

doringeman:
```
make docker-run-vllm-rocm
```

```
MODEL_RUNNER_HOST=http://localhost:8080 docker model smollm2-vllm
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -111,6 +111,27 @@ RUN curl -LsSf https://astral.sh/uv/install.sh | sh \
 
 RUN /opt/vllm-env/bin/python -c "import vllm; print(vllm.__version__)" > /opt/vllm-env/version
 
+# --- vLLM ROCm variant ---
+FROM llamacpp AS vllm-rocm
+
+ARG VLLM_VERSION=0.15.1
+
+USER root
+
+RUN apt update && apt install -y python3 python3-venv python3-dev curl ca-certificates build-essential && rm -rf /var/lib/apt/lists/*
+
+RUN mkdir -p /opt/vllm-env && chown -R modelrunner:modelrunner /opt/vllm-env
+
+USER modelrunner
+
+# Install uv and vLLM with ROCm support
+RUN curl -LsSf https://astral.sh/uv/install.sh | sh \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -111,6 +111,27 @@ RUN curl -LsSf https://astral.sh/uv/install.sh | sh \
...
+
+USER modelrunner
+
+# Install uv and vLLM with ROCm support
+RUN curl -LsSf https://astral.sh/uv/install.sh | sh \
+    && ~/.local/bin/uv venv --python /usr/bin/python3 --system-site-packages /opt/vllm-env \
+    && ~/.local/bin/uv pip install --python /opt/vllm-env/bin/python \
+    vllm==${VLLM_VERSION} --extra-index-url https://wheels.vllm.ai/rocm/
+
+RUN /opt/vllm-env/bin/python -c "import vllm; print(vllm.__version__)" > /opt/vllm-env/version
```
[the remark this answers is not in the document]
ericcurtin: It's a fair point about the duplication here

Makefile
```diff
@@ -5,8 +5,10 @@ LLAMA_SERVER_VERSION := latest
 LLAMA_SERVER_VARIANT := cpu
 BASE_IMAGE := ubuntu:24.04
 VLLM_BASE_IMAGE := nvidia/cuda:13.0.2-runtime-ubuntu24.04
+VLLM_ROCM_BASE_IMAGE := rocm/pytorch:rocm6.3.4_ubuntu24.04_py3.12_pytorch_release_2.6.0
```
ericcurtin, line 8: vllm tends to pull it's own version of everything on install in the venv, etc. We might be able to use plain-old ubuntu for all vLLM images. Llama.cpp is a bit different, need to install the dependancies ourselves

ericcurtin: llama.cpp and vLLM can be on different ROCm versions if needs be. Hard to avoid that, we are moving to the one container image per backend model anyway.

ericcurtin:
They now have a neat website which tells us the recommended way to install vLLM for the various platforms: https://vllm.ai/

PR #640, opened 2026-02-05
fix(model): restore implicit defaults for model identifiers

### LINKED ISSUE #636: Model ID naming issues
If I run `docker model list`, for **mxbai-embed** I get:
```
mxbai-embed-large
```

If I try to use it (in go), with:

`mxbai-embed-large`: `model not available on the specified engine URL`
`ai/mxbai-embed-large`: `model not available on the specified engine URL`
`mxbai-embed-large:latest`: `model not available on the specified engine URL`
**`ai/mxbai-embed-large:latest`: ✅**

Pnkcaht:
## What is changed
- Normalize shorthand model identifiers in `docker model run`, restoring implicit defaults and fixing model resolution.

## Related issue
- Fixes #636

## Diagram 

```mermaid
flowchart LR
    A[docker model list] -->|mxbai-embed-large| B[docker model run]
    B --> C[normalizeModelIdentifier]
    C --> D[ai/mxbai-embed-large:latest]
    D --> E[Model Runner Engine]
```

### DIFF
--- a/cmd/cli/commands/utils.go
+++ b/cmd/cli/commands/utils.go
@@ -172,7 +172,7 @@ func normalizeModelIdentifier(model string) string {
 	}
 
 	// If no tag is provided, assume the default tag.
-	if !strings.Contains(model, ":") {
+if lastSlash := strings.LastIndex(model, "/"); !strings.Contains(model[lastSlash+1:], ":") {
 		model = model + ":" + defaultTag
 	}
 


ericcurtin:
We have been trying to do all normalisation daemon-side rather than client side. Can that be done here?

ilopezluna:
@Pnkcaht as @ericcurtin mentioned the normalization should happen in the daemon-side, actually its already there: https://github.com/docker/model-runner/blob/9acf1ed9148024814744306f9522055d1db8a283/pkg/distribution/distribution/client.go#L116 (note is the distribution client which is called in the http handler through model manager).

For future contributions to fix specific issues like this one, please provide a test that reproduces the issue 🙏 this way we can easily validate your contribution.

ghost:
> Thanks for the contribution @Pnkcaht I have the same ask as in [#641 (review)](https://github.com/docker/model-runner/pull/641#pullrequestreview-3773459890) could you add a test that reproduces the issue with the current normalization we have in the backend?

I attempted to reproduce the issue via unit tests but couldn’t fully reproduce the reported behavior.
This change normalizes model identifiers on the CLI side to make the behavior more consistent when reusing values returned by docker model list. Sorry for this friend @ilopezluna

ilopezluna:
> > Thanks for the contribution @Pnkcaht I have the same ask as in [#641 (review)](https://github.com/docker/model-runner/pull/641#pullrequestreview-3773459890) could you add a test that reproduces the issue with the current normalization we have in the backend?
> 
> I attempted to reproduce the issue via unit tests but couldn’t fully reproduce the reported behavior. This change normalizes model identifiers on the CLI side to make the behavior more consistent when reusing values returned by docker model list. Sorry for this friend @ilopezluna

Thanks for trying to reproduce the issue via tests @Pnkcaht 👏 
Closing the PR as there is no issue to be fixed after all :)

PR #641, opened 2026-02-05
fix: avoid re-downloading models already present in local store

### LINKED ISSUE #634: Models appear in `docker model list`, but `docker model run` re-downloads
I noticed this issue where some of my models that appear in `docker model list` are not recognized by `docker model run` and it tries to re-download them

OS: Ubuntu 25.10

```shell
$ docker model version
Docker Model Runner version v1.0.2-99-gca12aae1-dirty
Docker Engine Kind: Docker Engine
```

```shell
$ docker model status
Docker Model Runner is running

Status:
mlx: not installed
sglang: python3 not found in PATH
vllm: vLLM binary not found
diffusers: python3 not found in PATH
llama.cpp: running llama.cpp version: c55bce4
```

### Screenshot

Image

Pnkcaht:
## What I did
- Stop re-downloading models already present in the local store by resolving cache via normalized model references.

## Related issue
- Fixes #634

## Diagram

```mermaid
flowchart LR
    A[docker model run] --> B[normalize model ref]
    B --> C{model in local store?}
    C -->|yes| D[use cached model]
    C -->|no| E[pull from registry]
```

### DIFF
--- a/pkg/distribution/distribution/client.go
+++ b/pkg/distribution/distribution/client.go
@@ -239,25 +239,27 @@ func (c *Client) PullModel(ctx context.Context, reference string, progressWriter
 	// Normalize the model reference
 	reference = c.normalizeModelName(reference)
 
-	// Fast-path: if the model already exists locally by normalized reference,
-	// do not hit the registry or re-download.
-	if localModel, err := c.store.Read(reference); err == nil {
-		c.log.Infoln("Model found in local store by tag:", utils.SanitizeForLog(reference))
+	// Fast-path only for immutable digest references.
+	// Tags may be mutable and must be checked against the registry.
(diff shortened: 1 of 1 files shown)


ericcurtin:
While this code looks ok. @ilopezluna said it's not necessary because of:

https://github.com/docker/model-runner/pull/617

If @ilopezluna is right and this doesn't provide any extra value, we should probably leave unmerged. But thanks for the efforts @Pnkcaht .

A test included here might have solved all doubts.

ilopezluna:
@Pnkcaht I'm closing the PR as it seems not to be needed. If I'm wrong, please feel free to reopen it with a test that reproduces the issue 🙏

PR #645, opened 2026-02-09
feat: migrate legacy hf.co tags to huggingface.co during client initialization

ilopezluna:
Models pulled with `hf.co/` before the normalization fix in #617 were stored with un-normalized tags, causing cache misses and duplicate downloads. This adds an automatic tag migration on client startup that rewrites `hf.co/` → `huggingface.co/` in the store index, so previously-pulled models are found by the cache check.


doringeman (approved):
Chatted on Zoom, the above is a different issue.
LGTM!

ilopezluna:
> Chatted on Zoom, the above is a different issue. LGTM!

@doringeman Added new issue found [here](https://github.com/docker/model-runner/issues/647)

PR #652, opened 2026-02-10
feat: make vllm-metal backend installation opt-in

doringeman:
This is needed as people might not want to have `vllm-metal` installed by default, especially for Docker Desktop users.
Eventually we'll switch to an approach when all the backends will be so self-contained and independently installable.

To test:
```
make -C cmd/cli install
```

```
rm -rf ~/.docker/model-runner/vllm-metal/ # make sure vllm-metal is not installed
MODEL_RUNNER_PORT=8080 make run
```

```
$ MODEL_RUNNER_HOST=http://localhost:8080 docker model run huggingface.co/mlx-community/llama-3.2-1b-instruct-4bit hi
Failed to generate a response: error response: status=412 body=backend "vllm-metal" is not installed; run: docker model install-runner --backend vllm-metal

$ MODEL_RUNNER_HOST=http://localhost:8080 docker model install-runner --backend vllm-metal
Installing vllm-metal backend...
vllm-metal backend installed successfully

$ MODEL_RUNNER_HOST=http://localhost:8080 docker model run huggingface.co/mlx-community/llama-3.2-1b-instruct-4bit hi
Hi there! How can I help you today?

$ curl -X POST http://localhost:8080/v1/chat/completions \
    -d '{
      "model": "hf.co/mlx-community/llama-3.2-1b-instruct-4bit",
      "messages": [{"role": "user", "content": "hi"}],
      "max_tokens": 512
    }'
{"id":"chatcmpl-06e20393","object":"chat.completion","created":1770745785,"model":"hf.co/mlx-community/llama-3.2-1b-instruct-4bit","choices":[{"index":0,"message":{"role":"assistant","content":"Hi there! How can I help you today?"},"finish_reason":"stop"}],"usage":{"prompt_tokens":3,"completion_tokens":8,"total_tokens":11}}
```

### DIFF
--- a/pkg/inference/scheduling/http_handler.go
+++ b/pkg/inference/scheduling/http_handler.go
@@ -198,11 +198,28 @@ func (h *HTTPHandler) handleOpenAIInference(w http.ResponseWriter, r *http.Reque
 		backend = h.scheduler.selectBackendForModel(model, backend, request.Model)
 	}
 
+	// If a deferred backend needs on-demand installation and the request
+	// comes from the model CLI, stream progress messages so the user sees
+	// what is happening while the download runs.
+	autoInstall := h.scheduler.installer.deferredBackends[backend.Name()] &&
+		!h.scheduler.installer.isInstalled(backend.Name()) &&
+		strings.Contains(r.UserAgent(), modelCLIUserAgentPrefix)
+	if autoInstall {
+		fmt.Fprintf(w, "Installing %s backend...\n", backend.Name())
+		if f, ok := w.(http.Flusher); ok {
+			f.Flush()
+		}
+	}
+
 	// Wait for the corresponding backend installation to complete or fail. We
 	// don't allow any requests to be scheduled for a backend until it has
 	// completed installation.
 	if err := h.scheduler.installer.wait(r.Context(), backend.Name()); err != nil {
-		if errors.Is(err, ErrBackendNotFound) {
+		if autoInstall {
+			// Headers are already sent (200 OK) from the progress
+			// line, so we can only write the error as plain text.
+			fmt.Fprintf(w, "backend installation failed: %v\n", err)
+		} else if errors.Is(err, ErrBackendNotFound) {
 			http.Error(w, err.Error(), http.StatusNotFound)
 		} else if errors.Is(err, errInstallerNotStarted) {
 			http.Error(w, err.Error(), http.StatusServiceUnavailable)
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -208,6 +208,12 @@ func main() {
 		backends[vllmmetal.Name] = vllmMetalBackend
 	}
 
+	// Backends whose installation is deferred until explicitly requested.
+	var deferredBackends []string
```
ilopezluna, line 212: I like this `defferedBackends` idea 👏

pkg/inference/scheduling/http_handler.go
```diff
@@ -97,6 +97,7 @@ func (h *HTTPHandler) routeHandlers() map[string]http.HandlerFunc {
 	m["GET "+inference.InferencePrefix+"/v1/models"] = h.handleModels
 	m["GET "+inference.InferencePrefix+"/v1/models/{name...}"] = h.handleModels
 
+	m["POST "+inference.InferencePrefix+"/install-backend"] = h.InstallBackend
```
ilopezluna, line 100: Should we use only a `POST` request to `/engines` to make the API more RESTful, or keep verbs in the path like we do for actions such as ps and df?

ericcurtin:
I'd rather we auto-pulled it on-demand tbh, I want the huggingface help for example to be only this:

```
docker model run huggingface.co/mlx-community/llama-3.2-1b-instruct-4bit
```

If a user wants to Ctrl-C that's fine, we should auto-pull and install like a "docker run" . If we have to change the huggingface help from a single "docker run" to multiple DMR commands, we have failed.

pkg/inference/scheduling/http_handler.go
doringeman: We could... do you want to do it in this PR? Or start an initiative for re-reviewing the whole API?

doringeman:
@ericcurtin thanks, done in in https://github.com/docker/model-runner/pull/652/commits/a5efa9f4c589084ed754acce9aaf663279c37763 and https://github.com/docker/model-runner/pull/652/commits/8b6c1ae1dba25ff6a102d818c9fb40a2650d7bb0.
```
$ MODEL_RUNNER_HOST=http://localhost:8080 docker model run huggingface.co/mlx-community/llama-3.2-1b-instruct-4bit hi
Installing vllm-metal backend...
vllm-metal backend installed successfully
Hi there! How can I help? What would you like to talk about?
```

pkg/inference/scheduling/http_handler.go
ilopezluna: Oh, sorry for not being clear. No need to change anything, I was just wondering what would be better since we follow both patterns. I usually prefer a RESTful API, but I’m also perfectly fine using verbs in the URL.

PR #654, opened 2026-02-11
Fix non-deterministic OCI digests from time.Now() in creation timestamp

### LINKED ISSUE #647: Hugging Face models produce non-deterministic OCI digests due to `time.Now()` in creation date
### Problem

When pulling a model from Hugging Face, we build the OCI artifact locally (since Hugging Face is not an OCI registry). The creation date in the OCI config is set to `time.Now()`, which means the same model pulled at different points in time produces a different config JSON and therefore a __different SHA digest__.

This breaks content-addressability: the same logical model (same weights, same config files, same revision) should always produce the same OCI digest regardless of when it was pulled.

__Affected code paths:__

- [`pkg/distribution/builder/builder.go` — `fromFormat()`](https://github.com/docker/model-runner/blob/main/pkg/distribution/builder/builder.go) — used when building from `FromPaths` (the Hugging Face pull path via `buildModelFromFiles`)

  ```go
  created := time.Now()  // ← non-deterministic
  ```

- [`pkg/distribution/builder/from_directory.go` — `FromDirectory()`](https://github.com/docker/model-runner/blob/main/pkg/distribution/builder/from_directory.go) — same pattern

  ```go
  created := time.Now()  // ← non-deterministic
  ```

Both set `types.Descriptor.Created` to the current wall-clock time, which is embedded in the OCI config JSON and influences the overall manifest digest.

### Impact

- Pulling the same HF model twice yields different digests, making it appear as two distinct models in the local store.
- Digest-based deduplication and caching cannot work correctly.
- Users cannot reliably pin or verify a model by its digest.

### Proposed solutions

Use the Hugging Face last commit date for the revision__ Query the HuggingFace API (e.g., `GET /api/models/{repo}` which returns `lastModified`, or the commit-specific endpoint `GET /api/models/{repo}/revision/{revision}`) and use the commit timestamp as the `Created` date. This is semantically correct—the model was "created" when it was last committed—and is deterministic for a given revision.

### Additional context

The `RepoFile` struct in `pkg/distribution/huggingface/repository.go` does not currently carry any timestamp from Hugging Face. We would need to either:

1. Fetch the repo/revision metadata (a single API call) to get the commit date, or
2. Propagate a `createdAt` parameter through `BuildModel` → `buildModelFromFiles` → `builder.FromPaths`/`fromFormat`.

veeceey:
Pulling the same HuggingFace model at different times produces different OCI digests because `time.Now()` is used as the creation timestamp in the config. This makes content-addressability unreliable since identical model content yields different manifests.

## Changes

- **Builder**: Add optional `WithCreated(time.Time)` build option to `FromPath`/`FromPaths` so callers can supply a deterministic timestamp. Uses `time.Now()` as fallback when not specified, keeping backward compatibility via variadic params.
- **FromDirectory**: Add `WithCreatedTime(time.Time)` directory option for the same purpose.
- **HuggingFace client**: Add `GetRepoInfo()` to fetch repo metadata including `lastModified`.
- **HuggingFace model builder**: Fetch `lastModified` from the HF API after downloading files and pass it as the creation timestamp. Falls back gracefully to `time.Now()` if the API call fails.

## Test plan

- [x] New tests verify same file + same timestamp = identical digest
- [x] New tests verify different timestamps = different digests
- [x] New tests for `GetRepoInfo` (success, default revision, 404)
- [x] All existing tests pass (`go test ./pkg/distribution/...`)

Fixes #647

### DIFF
--- a/pkg/distribution/huggingface/client.go
+++ b/pkg/distribution/huggingface/client.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"path"
 	"strings"
 	"time"
@@ -188,9 +189,9 @@ func (c *Client) GetRepoInfo(ctx context.Context, repo, revision string) (*RepoI
 		revision = "main"
 	}
 
-	url := fmt.Sprintf("%s/api/models/%s/revision/%s", c.baseURL, repo, revision)
+	reqURL := fmt.Sprintf("%s/api/models/%s/revision/%s", c.baseURL, repo, url.PathEscape(revision))
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, http.NoBody)
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, reqURL, http.NoBody)
 	if err != nil {
 		return nil, fmt.Errorf("create request: %w", err)
 	}
--- a/pkg/distribution/huggingface/model.go
+++ b/pkg/distribution/huggingface/model.go
@@ -80,7 +80,9 @@ func BuildModel(ctx context.Context, client *Client, repo, revision, tag string,
 	// produces the same OCI digest regardless of when it was pulled.
 	var createdTime *time.Time
 	repoInfo, err := client.GetRepoInfo(ctx, repo, revision)
-	if err == nil && !repoInfo.LastModified.IsZero() {
+	if err != nil {
+		log.Printf("Warning: could not fetch repo info for deterministic timestamp: %v. Falling back to current time.", err)
+	} else if !repoInfo.LastModified.IsZero() {
 		createdTime = &repoInfo.LastModified
 	}
 


veeceey:
### Manual test results

All distribution tests pass:

```
$ go test ./pkg/distribution/... -count=1
ok  	github.com/docker/model-runner/pkg/distribution/builder           0.190s
ok  	github.com/docker/model-runner/pkg/distribution/distribution      4.204s
ok  	github.com/docker/model-runner/pkg/distribution/files             0.353s
ok  	github.com/docker/model-runner/pkg/distribution/format            1.189s
ok  	github.com/docker/model-runner/pkg/distribution/huggingface       0.797s
ok  	github.com/docker/model-runner/pkg/distribution/internal/bundle   0.587s
ok  	github.com/docker/model-runner/pkg/distribution/internal/gguf     1.412s
ok  	github.com/docker/model-runner/pkg/distribution/internal/mutate   0.991s
ok  	github.com/docker/model-runner/pkg/distribution/internal/partial  2.453s
ok  	github.com/docker/model-runner/pkg/distribution/internal/progress 0.868s
ok  	github.com/docker/model-runner/pkg/distribution/internal/safetensors 2.659s
ok  	github.com/docker/model-runner/pkg/distribution/internal/store    2.930s
ok  	github.com/docker/model-runner/pkg/distribution/modelpack         1.754s
ok  	github.com/docker/model-runner/pkg/distribution/oci/authn         1.572s
ok  	github.com/docker/model-runner/pkg/distribution/packaging         1.990s
ok  	github.com/docker/model-runner/pkg/distribution/registry          2.601s
ok  	github.com/docker/model-runner/pkg/distribution/tarball           2.596s
```

New test details:

- `TestWithCreatedDeterministicDigest`: builds the same GGUF file twice with an identical fixed timestamp, verifies digests match; then builds with a different timestamp and verifies the digest changes.
- `TestWithCreatedFromPaths`: same as above but through `FromPaths`.
- `TestFromDirectoryWithCreatedTime`: verifies `WithCreatedTime` produces deterministic digests from `FromDirectory`.
- `TestClientGetRepoInfo`: mock server returns `lastModified` JSON, verifies parsed timestamp.
- `TestClientGetRepoInfoDefaultRevision`: verifies empty revision defaults to "main".
- `TestClientGetRepoInfoNotFound`: verifies `NotFoundError` on 404.

Backward compatibility: `FromPath`, `FromPaths`, and `FromDirectory` all use variadic optional params, so existing callers are unaffected.

ericcurtin:
This looks ok, so far. Could you go through the Gemini comments @veeceey ?

veeceey:
Thanks @ericcurtin! Addressed the Gemini and Sourcery review comments:

1. **URL-escape revision**: The revision parameter in `GetRepoInfo` is now URL-escaped with `url.PathEscape()` so values like `refs/pr/1` are handled correctly. (The existing `ListFiles` and `DownloadFile` methods have the same pattern -- I can fix those too in a follow-up if you'd like.)

2. **Log warning on GetRepoInfo failure**: Instead of silently falling back to `time.Now()`, we now log a warning when `GetRepoInfo` fails, so users can see why they might not get deterministic digests.

3. **WithCreatedTime vs WithCreated naming**: Kept `WithCreatedTime` for the `DirectoryOption` since `WithCreated` already exists as a `BuildOption` in the same package -- renaming it would cause a compilation error. The different names disambiguate between the two option types.

ilopezluna:
Thank you @veeceey while testing this PR I noticed we have (at least) another non-deterministic field that causes different manifest hash: `ModTime` in layer annotations, we use the downloaded time for that field as well. I'm going to open another issue for that 👍

ilopezluna:
This is the follow up issue: https://github.com/docker/model-runner/issues/669 in case you are interested @veeceey its all yours :)

veeceey:
I’ll take a look at it and fix :)

________________________________

veeceey:
Nice catch on the ModTime in layer annotations! I'll pick up #669 — should be a straightforward extension of the same approach, passing the timestamp through to the annotation layer as well.

PR #661, opened 2026-02-11
improve launch command

ericcurtin:
- Allow `docker model launch` with no args to list supported apps with descriptions and install status
- Add `--config` flag to print configuration without launching
- Properly handle `--` separator for passing extra args to integrations
- Reject unexpected extra args without `--` separator with helpful error
- Add app descriptions for all supported apps
- Update documentation and tests

### DIFF
--- a/cmd/cli/commands/launch.go
+++ b/cmd/cli/commands/launch.go
@@ -78,24 +78,63 @@ var supportedApps = func() []string {
 	return apps
 }()
 
+// appDescriptions provides human-readable descriptions for supported apps.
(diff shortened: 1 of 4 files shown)


ilopezluna:
I think its worth to add the Sourcery suggestion:

> Consider requiring dashIdx == 1 and erroring otherwise, or clearly documenting that only args after -- are forwarded and earlier ones are ignored.

ericcurtin:
Addressed comments @ilopezluna

PR #679, opened 2026-02-16
ci: add llama.cpp bump script and scheduled workflow

doringeman:
E.g.,

```
$ ./scripts/bump-llamacpp.sh
Fetching latest tags from llama.cpp...
Current: b7923 (c55bce415926c7e8484f20d854afebe0168f7513)
Target:  b8068 (267ba5a1d957158316a4fc85b1f0cf316d9a5233)

Updating submodule to b8068...
Staging submodule change...

Done. Commit with:
  git commit --signoff -S -m 'chore: bump llama.cpp (https://github.com/ggml-org/llama.cpp/releases/b8068)'
```

The workflow runs the script every Monday or on manual dispatch, and opens a PR with the update.

In a follow-up PR I'll add some build and test steps to verify the new llama.cpp compiles and works as expected.

### DIFF
--- a/.github/workflows/bump-llamacpp.yml
+++ b/.github/workflows/bump-llamacpp.yml
@@ -1,0 +1,55 @@
+name: Bump llama.cpp
+
+permissions:
+  actions: write
+  contents: write
+  pull-requests: write
+
+on:
+  schedule:
+    - cron: '0 5 * * 1'
+  workflow_dispatch:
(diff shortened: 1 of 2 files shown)


ericcurtin (approved):
Not great for stability, but I agree the trade off is probably still worth it right now to move fast and automatically do this.

PR #705, opened 2026-02-20
implement manual approval for Docker CE release workflow

ilopezluna:
### Problem
The release workflow was blocking while waiting for the `docker/release-repo` workflow to complete. The external workflow requires manual approval at the `deploy-to-live` gate, which could take hours to be approved. During this wait time, the entire `release-cli-docker-ce` job would be stuck, consuming GitHub Actions minutes and blocking other dependent jobs.

### Solution
Split the `release-cli-docker-ce` job into two separate jobs with a manual approval gate:

1. **`release-cli-docker-ce-trigger`** - Triggers the packaging and release-repo workflows, then completes immediately
2. **`release-cli-docker-ce-wait`** - Requires manual approval via GitHub environment protection before waiting for the release-repo workflow to complete

I also removed the timeouts, I think they are unnecessary, I will get back any of them if turns out they are needed

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -604,46 +604,19 @@ jobs:
           echo "✅ Packaging workflow completed successfully"
 
       - name: Trigger deploy workflow
-        id: trigger_release_repo
         env:
           GH_TOKEN: ${{ secrets.CLI_RELEASE_PAT }}
           VERSION: ${{ needs.prepare.outputs.version }}
           PACKAGING_IMAGE: ${{ steps.packaging.outputs.packaging_image }}
         run: |
           echo "🚀 Triggering deploy workflow"
-
-          # gh workflow run returns run URL on stdout (gh v2.87.0+)
-          OUTPUT=$(gh workflow run plugin.yml \
+          gh workflow run plugin.yml \
             --repo docker/release-repo \
             --ref production \
             -f packaging_image="$PACKAGING_IMAGE" \
             -f model_version="$VERSION" \
             -f channel=stable \
-            -f release_live=true)
-
-          RUN_URL=$(echo "$OUTPUT" \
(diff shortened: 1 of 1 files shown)


ericcurtin (approved):

.github/workflows/release.yml
```diff
@@ -640,10 +648,49 @@ jobs:
             -f release_live=true
           echo "✅ Plugin release workflow triggered in docker/release-repo"
 
+      - name: Post trigger summary
```
doringeman: `gh workflow run` has just been updated to return the workflow ID!

https://github.com/cli/cli/issues/4001#issuecomment-3932621537
❤️ https://github.blog/changelog/2026-02-19-workflow-dispatch-api-now-returns-run-ids/

babakks: Hi from `gh`! :wave: Sorry to interrupt. I just noticed this PR via links and thought sharing these:

- When run `gh workflow run` non-interactively (as in here, in Actions), the dispatched run URL is the only thing that is printed on stdout. So you can safely drop stderr redirection (`2>&1`) to avoid any accidental matches against error messages if they happen to include URLs of the same pattern.

- Usually it takes a few weeks for new `gh` versions to be fully propagated across Actions runner images. For example, you can watch for changes on `ubuntu-latest` image [here](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#cli-tools).

Cheers

doringeman: Hi @babakks! 👋 Nice, thanks for the pointers! 😁

ilopezluna: Thank you both! update in [55189ae](https://github.com/docker/model-runner/pull/705/commits/55189ae62b3c2907ca5158eb2d3dd2d29cd142da)

PR #717, opened 2026-02-27
set ContentLength when forwarding requests

ericcurtin:
HTTP/2 clients (e.g. Java HttpClient with HTTP_2 version) often omit the Content-Length header since HTTP/2 uses DATA frames for body framing. When DMR's reverse proxy forwards such requests to the backend via HTTP/1.1, it uses Transfer-Encoding: chunked (ContentLength == -1), which vLLM's Python/uvicorn server fails to parse — resulting in an empty body and a 422 Unprocessable Entity response.

Fix by explicitly setting ContentLength = len(body) on the upstream request after replacing the body with the already-buffered bytes. This ensures a Content-Length header is always sent, consistent with how the Ollama and Anthropic handlers already handle this. llama.cpp was unaffected because its C/C++ HTTP server handles chunked encoding gracefully.

### DIFF
--- a/pkg/inference/scheduling/http_handler.go
+++ b/pkg/inference/scheduling/http_handler.go
@@ -271,8 +271,14 @@ func (h *HTTPHandler) handleOpenAIInference(w http.ResponseWriter, r *http.Reque
 	}()
 
 	// Create a request with the body replaced for forwarding upstream.
+	// Set ContentLength explicitly so the backend always receives a Content-Length
+	// header. Without this, HTTP/2 requests (where clients may omit Content-Length)
+	// are forwarded with Transfer-Encoding: chunked, which some backends (e.g.
+	// vLLM's Python/uvicorn server) fail to parse, resulting in an empty body and
+	// a 422 response.
 	upstreamRequest := r.Clone(r.Context())
 	upstreamRequest.Body = io.NopCloser(bytes.NewReader(body))
+	upstreamRequest.ContentLength = int64(len(body))
 
 	// Perform the request.
 	runner.ServeHTTP(w, upstreamRequest)


dandyfopper:
@ericcurtin - Tested on macOS (Apple Silicon) with the patched binary (v1.1.8-2-g771b9b0a) running standalone on port 13434.

The standalone model-runner binary doesn't appear to support h2c, so I wasn't able to exercise the HTTP/2 code path directly:

curl --http2 sends HTTP/1.1 with an Upgrade: h2c header — the server doesn't upgrade, and the request still hits vLLM as HTTP/1.1 (422)
curl --http2-prior-knowledge fails immediately — "peer does not support HTTP/2 properly"
curl --http1.1 works fine (200 OK, as expected)
It seems like the h2c handling only comes into play inside Docker Desktop's networking/proxy layer, not in the standalone binary. Could you advise on how to test this? Specifically:

Is there a way to swap the patched binary into Docker Desktop's model-runner?
Or does Docker Desktop's proxy handle the h2c → HTTP/1.1 translation before it reaches model-runner, meaning the fix needs to be tested inside that stack?
Happy to re-test however you recommend.

Is there a plan to support streaming as well?

ericcurtin:
Streaming in what way? (sometimes that word in used in different ways in this space)

ilopezluna (approved):

dandyfopper:
By streaming I mean SSE (Server-Sent Events) — when I send "stream": true in the request body:

{
  "model": "ai/gemma3-vllm:latest",
  "stream": true,
  "messages": [{"role": "user", "content": "Say hello"}]
}

I'd expect the response to come back as Content-Type: text/event-stream with incremental data: {"object":"chat.completion.chunk",...} lines, matching the [OpenAI streaming spec](https://platform.openai.com/docs/api-reference/chat/create#chat-create-stream).

Instead, vLLM returns a single Content-Type: application/json response with "object": "chat.completion" (non-chunked) — the stream parameter is effectively ignored.

This is a separate issue from the HTTP/2 body-dropping bug — happy to split it into its own issue if that's cleaner.

PR #721, opened 2026-03-02
feat: expose richer model metadata in v1/models

### LINKED ISSUE #143: Enhancement: Expose richer model metadata in /engines/v1/models (similar to OpenRouter)
Hello team,

I would like to propose an enhancement to the Model Runner API.

**Current behavior**  
The `/engines/v1/models` endpoint currently returns only a minimal schema for each model (e.g., `id`, `object`, `created`, `owned_by`).

**Proposed enhancement**  
Consider expanding the model schema to include additional metadata about each model — similar to what OpenRouter’s Models API exposes. OpenRouter’s API provides comprehensive metadata for each model in its standard response, including a human-readable name, description, context length, technical architecture (input/output modalities, tokenizer, instruct_type), pricing information and supported parameters. This metadata allows client applications and agent frameworks to understand a model’s capabilities and limitations without hard-coding them. Their documentation emphasises that the API is designed to make important information about LLMs freely available for integration.

**Why it matters**  
Applications like [E‑Worker Studio](https://app.eworker.ca) dynamically adjust UI capabilities based on the available model features. When only a model ID is provided, it’s difficult to know what modalities are supported (text, image, audio), maximum context window, or any special parameters. By exposing richer metadata, Model Runner could enable downstream tools to present more accurate options to users and better utilise models’ full capabilities. It also aligns the Model Runner API with emerging standards in the ecosystem.

VedantMadane:
Fixes #143.

This PR expands the model schema returned by the \/v1/models\ endpoint to include additional metadata about each model, similar to what other providers like OpenRouter expose.

**New fields added to OpenAIModel:**
- \context_window\: Maximum context length supported by the model.
- \rchitecture\: Technical architecture of the model.
- \parameters\: Description of model parameters.
- \quantization\: Quantization method used.
- \size\: Parameter size (e.g., " 8B\).

### DIFF
--- a/pkg/inference/models/api_test.go
+++ b/pkg/inference/models/api_test.go
@@ -320,16 +320,16 @@ type mockModel struct {
 	desc   types.Descriptor
 }
 
-func (m *mockModel) ID() (string, error)                    { return m.id, nil }
-func (m *mockModel) Tags() []string                         { return m.tags }
-func (m *mockModel) Config() (types.ModelConfig, error)     { return m.config, nil }
-func (m *mockModel) Descriptor() (types.Descriptor, error)  { return m.desc, nil }
-func (m *mockModel) GGUFPaths() ([]string, error)           { return nil, nil }
-func (m *mockModel) SafetensorsPaths() ([]string, error)    { return nil, nil }
-func (m *mockModel) DDUFPaths() ([]string, error)           { return nil, nil }
-func (m *mockModel) ConfigArchivePath() (string, error)     { return "", nil }
-func (m *mockModel) MMPROJPath() (string, error)            { return "", nil }
-func (m *mockModel) ChatTemplatePath() (string, error)      { return "", nil }
+func (m *mockModel) ID() (string, error)                   { return m.id, nil }
+func (m *mockModel) Tags() []string                        { return m.tags }
+func (m *mockModel) Config() (types.ModelConfig, error)    { return m.config, nil }
+func (m *mockModel) Descriptor() (types.Descriptor, error) { return m.desc, nil }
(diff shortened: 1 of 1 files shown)


doringeman:
Hey @VedantMadane, why do you need this?
This API you've modified corresponds with the official [List models](https://developers.openai.com/api/reference/resources/models/methods/list).
Isn't [/models](https://github.com/docker/model-runner/blob/3561e7961e632209e347b06a76374313baecbf50/pkg/inference/models/http_handler.go#L177) what you're looking for? E.g., `curl -s http://localhost:12434/models | jq .`

VedantMadane:
Hi @doringeman, good question! The `/engines/llama.cpp/v1/models` endpoint (OpenAI-compatible List Models) currently returns only `id` and `object` fields, while the `/models` endpoint returns more details but in a non-standard format.

This PR enriches the OpenAI-compatible response with metadata fields like `architecture`, `parameters`, `quantization`, `context_window`, and `size` — so consumers using the standard OpenAI SDK can access model capabilities without needing a separate non-standard call.

The idea is to keep the interface OpenAI-compatible while exposing the richer metadata that Docker Model Runner already has.

ericcurtin:
Missing unit tests for new fields, ContextWindow should be *int32 for proper omitempty behavior.

doringeman:
Hey @VedantMadane , WDYT about adding all the new extra fields under a common field called `"dmr"`?
E.g.,
```
{
  "id": "VAR_chat_model_id",
  "object": "model",
  "created": 1686935002,
  "owned_by": "openai",
  "dmr": {
     "custom-field":"custom-vaalue"
   }
}
```

PR #738, opened 2026-03-07
cli: prompt and install missing backend during model run

iamanishx:
## Summary
- detect the required backend for `model run` before pulling the model by inspecting local metadata first and falling back to remote metadata
- prompt the user to confirm backend installation when the required backend is missing, then run `install-runner --backend <name>` automatically
- continue with pull and run only after backend availability is confirmed, and report a clear error if install did not actually make the backend available
- normalize backend status parsing so values like `running llama.cpp ...` are treated as installed while `not ...` and `error ...` states are treated as missing

## Why
Users running safetensors models had to run a separate backend install command manually. This change keeps users in the `model run` flow while still asking for explicit consent before downloading backend components.

## Validation
- `go build ./cmd/cli/...`
- `go test ./cmd/cli/commands/... -run "Install|Backend|Run"`
- manual validation in PowerShell with:
  - `./model-cli.exe --debug run ai/functiongemma-vllm:270M \"hello\"`
  - observed prompt: `Backend \"vllm\" is not installed. Download and install it now? [Y/n]:`


doringeman (commented):
Please see `func (c *Client) InstallBackend`, I believe that can be used to simplify the backend installation and the checks. 🙌

iamanishx:
> Please see `func (c *Client) InstallBackend`, I believe that can be used to simplify the backend installation and the checks. 🙌 

Will check this out thanks

iamanishx:
@doringeman i have made the changes please have a look

ericcurtin:
Fragile error string comparison, inconsistent constant usage (diffusers), missing import for diffusers package, no tests for new functions, unused GetRequiredBackend function, merge conflicts

cmd/cli/commands/utils.go
```diff
@@ -270,6 +278,105 @@ func newTable(w io.Writer) *tablewriter.Table {
...
+	}
+	if !installed {
+		return fmt.Errorf("backend %q is still not installed; run 'docker model install-runner --backend %s'", backend, backend)
+	}
+
+	cmd.Printf("Backend %q installed successfully.\n", backend)
+	return nil
+}
+
+func GetRequiredBackendFromModelInfo(modelInfo *dmrm.Model) (string, error) {
```
doringeman: Please check and use `func (s *Scheduler) selectBackendForModel`.
The server could expose a simple endpoint like:
```
GET /models/{name}/backend → {"backend": "vllm", "installed": true}
```

doringeman:
And sorry for the delay! 🙌

cmd/cli/commands/utils.go
doringeman: The backend selection logic in `GetRequiredBackendFromModelInfo` duplicates what the server already does in `Scheduler.selectBackendForModel`, and the two already diverge (the server has a platform-aware fallback chain: vLLM → MLX → SGLang, while the CLI always returns vllm for safetensors). Rather than maintaining this mapping in two places, consider calling the server to determine the required backend.

iamanishx:
> And sorry for the delay! 🙌

hey am really sorry for the delay i have pushed a commit please review

ilopezluna:
@sourcery-ai review

PR #756, opened 2026-03-17
downgrade Go version from 1.26.0 to 1.25.8

ilopezluna:
required to bump DMR in DD to latest version


thaJeztah:
Was the problem that `go.mod` required go1.26? If that's the case, then a good option is also to keep the version in  `go.mod` at an older version (generally "latest" and "latest-1" are supported by upstream Go), and for CI you can specify the version separately; also look at;

- https://github.com/docker/compose/pull/13418

doringeman:
Thanks! Yes, we downgraded so we can import it in projects that still use 1.25. I'm totally fine with using 1.25, too.

PR #771, opened 2026-03-20
gracefully fall back when model chat template doesn't support tools

ericcurtin:
Models like gemma3 use Jinja chat templates that only support alternating user/assistant roles and cannot handle the tool/assistant(tool_calls) message roles introduced by the agentic web-search loop.

When a 500 response containing "Jinja" is detected, the client now resets the message history to the original user prompt, clears the tool schemas, and retries — allowing the model to answer from training data rather than erroring out. This covers two failure modes:

  1. The model rejects the request immediately because the tools parameter causes its template to inject an incompatible system message.
  2. The model successfully emits a tool_call but then fails on the follow-up request because it can't process the resulting "tool" role message.

Adds three unit tests covering the fallback paths and the non-Jinja error case.

### DIFF
--- a/cmd/cli/desktop/desktop.go
+++ b/cmd/cli/desktop/desktop.go
@@ -24,10 +24,23 @@ import (
 	"go.opentelemetry.io/otel"
 )
 
+const (
+	// maxToolCallIterations caps the number of agentic tool-call rounds to
+	// prevent infinite loops when a model repeatedly requests tool calls.
+	maxToolCallIterations = 10
+)
+
 var (
 	ErrNotFound           = errors.New("model not found")
 	ErrServiceUnavailable = errors.New("service unavailable")
 )
+
+// ClientTool is a tool that can be registered with the chat client.
+type ClientTool interface {
+	Name() string
+	Schema() Tool
+	Execute(ctx context.Context, args map[string]any) (string, error)
+}
 
 type otelErrorSilencer struct{}
 
@@ -368,6 +381,13 @@ func (c *Client) Chat(model, prompt string, imageURLs []string, outputFunc func(
 	return c.ChatWithContext(context.Background(), model, prompt, imageURLs, outputFunc, shouldUseMarkdown)
 }
 
+// accumulatedToolCall collects streamed tool call fragments into a complete call.
+type accumulatedToolCall struct {
+	id        string
+	name      string
+	arguments strings.Builder
+}
+
 // Preload loads a model into memory without running inference.
 // The model stays loaded for the idle timeout period.
 func (c *Client) Preload(ctx context.Context, model string) error {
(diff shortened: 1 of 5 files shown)


cmd/cli/tools/websearch.go
```diff
@@ -0,0 +1,132 @@
...
+					},
+				},
+				"required": []string{"query"},
+			},
+		},
+	}
+}
+
+// Execute performs the web search using Exa's MCP endpoint.
+func (w *WebSearchTool) Execute(args map[string]any) (string, error) {
```
doringeman: I think it'd be good to propagate a context here as the user might want to cancel the request.

cmd/cli/desktop/desktop.go
```diff
@@ -29,6 +29,13 @@ var (
 	ErrServiceUnavailable = errors.New("service unavailable")
 )
 
+// ClientTool is a tool that can be registered with the chat client.
+type ClientTool interface {
+	Name() string
+	Schema() Tool
+	Execute(args map[string]any) (string, error)
+}
+
```
doringeman:
```suggestion
```
dead code

cmd/cli/desktop/desktop.go
```diff
@@ -484,162 +487,316 @@ func (c *Client) ChatWithMessagesContext(ctx context.Context, model string, conv
...
+			}
+
+			resp.Body.Close()
+			if err := scanner.Err(); err != nil {
+				return assistantResponse.String(), fmt.Errorf("error reading response stream: %w", err)
+			}
+		}
+
+		// If the model requested tool calls, execute them and loop.
+		if finishReason == "tool_calls" && len(pendingToolCalls) > 0 {
```
doringeman, line 757: Although models like `llama3.2` support tool calling, they're not good at them, neither calling sometimes nor interpreting the results. WDYT about adding a loop bound here - a global per all tools, or per tool?

cmd/cli/tools/websearch.go
```diff
@@ -0,0 +1,132 @@
...
+	query, ok := args["query"].(string)
+	if !ok || query == "" {
+		return "", fmt.Errorf("query parameter is required")
+	}
+
```
doringeman, line 55: FWIW I used a debug log here to test the tool-call loop.

cmd/cli/desktop/desktop.go
```diff
@@ -664,6 +821,12 @@ func (c *Client) ChatWithContext(ctx context.Context, model, prompt string, imag
 	return err
 }
 
+// ChatWithToolsContext is like ChatWithContext but supports tool calls.
+func (c *Client) ChatWithToolsContext(ctx context.Context, model, prompt string, imageURLs []string, outputFunc func(string), shouldUseMarkdown bool, tools ...ClientTool) error {
+	_, err := c.ChatWithMessagesContext(ctx, model, nil, prompt, imageURLs, outputFunc, shouldUseMarkdown, tools...)
+	return err
+}
+
```
doringeman:
```suggestion
```
dead code

cmd/cli/desktop/desktop.go
ericcurtin: Not worried about old models like llama3.2. llama4 (already released), llama5, etc. is going to make those models irrelevant in the not so distant future. If a model is bad at tool calling, well... It's gonna become irrelevant...

ericcurtin:
@doringeman addressed all comments

PR #782, opened 2026-03-24
Validate realm URL before token exchange

ilopezluna:
This pull request introduces important security enhancements to the `Exchange` function in the `pkg/distribution/oci/remote/transport.go` file to prevent Server-Side Request Forgery (SSRF) attacks, along with new tests to verify these protections. The main improvements include strict validation of token realm URLs, suppression of sensitive response bodies in error messages, and additional logging for failed token requests.

### DIFF
--- a/pkg/distribution/oci/remote/transport.go
+++ b/pkg/distribution/oci/remote/transport.go
@@ -163,27 +163,29 @@ func resolveAndValidateRealm(rawURL string) (dialAddr, hostname string, err erro
 // resolution. This closes the TOCTOU window between realm URL validation and
 // the actual connection. For TLS connections, serverName is set as the SNI
 // value so that certificate validation still uses the original hostname.
-func buildSafeTransport(base http.RoundTripper, dialAddr, serverName string) http.RoundTripper {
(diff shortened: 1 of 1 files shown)


pkg/distribution/oci/remote/transport.go
```diff
@@ -38,6 +41,153 @@ type Token struct {
...
+// successive lookups (TOCTOU).
+//
+// If hostname is a literal IP address it is validated directly without
+// triggering a DNS lookup.
+func resolveAndValidateRealm(rawURL string) (dialAddr, hostname string, err error) {
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return "", "", fmt.Errorf("invalid realm URL: %w", err)
+	}
+
```
doringeman, line 108:
```suggestion

	if u.Scheme != "https" && u.Scheme != "http" {
    	return "", "", fmt.Errorf("realm URL scheme %q is not allowed", u.Scheme)
	}

```
Otherwise URLs like `gopher://redis:6379/_SET%20key%20pwned` would bypass the validation, right?

PR #793, opened 2026-03-26
Add format dduf

ilopezluna:
This pull request updates the handling of model formats by introducing a new `FormatDDUF` type to replace the previously used `FormatDiffusers`, which is now deprecated but retained for backward compatibility.

Diffusers is not a format, so using it as `config.FormatDiffusers` is missleading.


ilopezluna:
> LGTM, with one question: shouldn't DDUF be handled in `func (s *Scheduler) selectBackendForModel` as well?

Yes! thanks, I'm going to add it in a follow-up PR

PR #794, opened 2026-03-26
Remove vLLM x86_64 NVIDIA platform warning from pull compat check

ericcurtin:


pkg/distribution/distribution/client.go
```diff
@@ -798,13 +797,6 @@ func checkCompat(image types.ModelArtifact, log *slog.Logger, reference string,
 
 	if config.GetFormat() == "" {
 		log.Warn("Model format field is empty for , unable to verify format compatibility", "model", utils.SanitizeForLog(reference))
-	} else if !slices.Contains(GetSupportedFormats(), config.GetFormat()) {
```
ilopezluna, line 801: `GetSupportedFormats` becomes dead code, you can remove it

ericcurtin:
@ilopezluna it really doesn't make sense, it appears on macOS even though we delivered vllm support there.

Also, I want people to report vllm doesn't work on my system and see the ugly vllm-log (just like they would see if they tried vanilla vllm). It gives us more feedback on where people are trying to run this.

ilopezluna:
> @ilopezluna it really doesn't make sense, it appears on macOS even though we delivered vllm support there.
> 
> Also, I want people to report vllm doesn't work on my system and see the ugly vllm-log (just like they would see if they tried vanilla vllm). It gives us more feedback on where people are trying to run this.


makes sense, thanks!

PR #797, opened 2026-03-26
Release as latest

ilopezluna:
This pull request removes the separate workflow for promoting Docker images to the `latest` tag and streamlines the release process by always tagging built images as `latest` within the main release workflow. The previous option to conditionally push the `latest` tag using the `pushLatest` input has been eliminated, and the logic for promoting images is now handled directly in the release workflow.


ericcurtin (approved):

doringeman (commented):
🤞

PR #853, opened 2026-04-10
fix: sandbox permissions

ilopezluna:
This pull request enhances the backend runner configuration and updates the macOS sandbox policy to better support the `vllm-metal` backend. The main changes include allowing custom environment variables for backend processes, updating the sandbox to permit necessary network and file operations, and setting a required environment variable for `vllm-metal`.

### DIFF
--- a/pkg/inference/backends/runner.go
+++ b/pkg/inference/backends/runner.go
@@ -42,6 +42,10 @@ type RunnerConfig struct {
 	// ErrorTransformer is an optional function to transform error output
 	// into a more user-friendly message. If nil, the raw output is used.
 	ErrorTransformer ErrorTransformer
+	// Env is an optional list of extra environment variables for the backend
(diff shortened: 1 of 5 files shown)


pkg/inference/backends/runner.go
```diff
@@ -88,6 +92,9 @@ func RunBackend(ctx context.Context, config RunnerConfig) error {
 			}
 			command.Stdout = config.ServerLogWriter
 			command.Stderr = out
+			if len(config.Env) > 0 {
+				command.Env = append(os.Environ(), config.Env...)
```
sathiraumesh, line 114: May be its good idea to parse the environment variable format as a defensive measure for the future, since we pass it as a string aka`"VLLM_HOST_IP=127.0.0.1"`.

ericcurtin (approved):

PR #873, opened 2026-04-20
fix: improve pull resume reliability and progress bar display on retry

ericcurtin:
- Only accept HTTP 206 with a matching Content-Range start offset as a
  successful Range response in rangeTransport; a 200 response means the
  server ignored the Range header and is sending from byte 0, so
  appending it to the partial file would corrupt the blob. A misbehaving
  server returning 206 with a different range is also rejected.

- Preserve .incomplete files on all read errors, not just context
  cancellation, so every kind of transient failure (network reset, stream
  error, etc.) can be resumed on the next attempt.

- Stop rolling back fully-downloaded layer blobs on Write failure.
  Layer blobs are content-addressed and immutable; keeping them lets a
  subsequent pull skip already-completed layers entirely instead of
  re-downloading them. The manifest, config, and index rollback still
  runs to leave the store in a consistent non-indexed state.

- Print a blank line to stdout before each retry so that orphaned
  progress bars from the failed attempt are visually separated from the
  new attempt's bars, preventing garbled terminal output.

- Increase Pull and Push max retries from 3 to 4 and update
  TestPullMaxRetriesExhausted to match (5 total attempts, error message
  says "after 4 retries").

### DIFF
--- a/cmd/cli/desktop/desktop.go
+++ b/cmd/cli/desktop/desktop.go
@@ -117,7 +117,7 @@ func (c *Client) Pull(model string, printer standalone.StatusPrinter) (string, b
 		hfToken = os.Getenv("HF_TOKEN")
 	}
 
-	return c.withRetries("download", 3, printer, func(attempt int) (string, bool, error, bool) {
+	return c.withRetries("download", 4, printer, func(attempt int) (string, bool, error, bool) {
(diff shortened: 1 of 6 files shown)


ilopezluna (approved):
I agree with Sourcery and Gemini comments but nothing blocking, we can add the suggestion in a follow up PR

PR #874, opened 2026-04-20
Fix responses goroutine leak

ilopezluna:
This pull request introduces proper resource cleanup for the router and service layers to prevent goroutine leaks, especially when the optional Responses API is enabled. The changes ensure that any background goroutines started by the router (such as those for the responses Store) are properly stopped when the service shuts down. It also adds tests to verify that resource cleanup is performed correctly and integrates a goroutine leak detector into the test suite.

### DIFF
--- a/main.go
+++ b/main.go
@@ -252,6 +252,7 @@ func main() {
 		log.Error("failed to initialize service", "error", err)
 		exitFunc(1)
 	}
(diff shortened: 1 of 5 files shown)


ilopezluna:
@sourcery-ai review

doringeman (approved):

PR #877, opened 2026-04-21
fix: preserve Range header across registry redirects for resumable pulls

ilopezluna:
Most registries (including Docker Hub) redirect blob downloads to a CDN via HTTP 302. The `rangeTransport` correctly set the Range header on the initial request, but Go's `http.Client` copies headers from the original request (before RoundTrip cloned and modified it) when following redirects. This caused the Range header to be lost on the redirect request to the CDN, which then served the full blob from byte 0.

Since `RangeSuccess` was never recorded, `WriteBlobWithResume` deleted the incomplete file and restarted the download from scratch on every resume attempt.

Fix: `rangeTransport.RoundTrip` now follows redirects at the transport level (up to 10 hops) when a Range header is set, ensuring the header is preserved across redirects to CDNs. Authorization headers are stripped on cross-domain redirects for security. Non-resume requests are completely unaffected.

### DIFF
--- a/pkg/distribution/oci/remote/range_redirect_test.go
+++ b/pkg/distribution/oci/remote/range_redirect_test.go
@@ -1,0 +1,537 @@
+package remote
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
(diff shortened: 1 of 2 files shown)


ilopezluna:
@sourcery-ai review

doringeman (approved):

PR #879, opened 2026-04-21
Use official llama.cpp images for Linux

ilopezluna:
Use official upstream images from [`ghcr.io/ggml-org/llama.cpp`](https://github.com/ggml-org/llama.cpp/pkgs/container/llama.cpp) for all Linux Docker images instead of building llama.cpp ourselves.

**Scope:** Linux only. macOS and Windows (Docker Desktop) are unchanged.

### Key changes
- Dockerfile pulls from upstream images (`server-vulkan`, `server-cuda13`, `server-rocm`)
- New `resolve-llama-upstream-image.sh` script maps variants to upstream GHCR tags with validation
- `resolveLlamaServerBin()` discovers `llama-server` (upstream) or falls back to `com.docker.llama-server` (Desktop)
- Build-time `ldd` check catches missing shared library dependencies early
- Added `libgomp1` to `apt-install.sh` (required by upstream CPU backend plugins)

### Follow-ups
- [ ] Add MUSA and OpenVINO variants ([comment](https://github.com/docker/model-runner/pull/879#discussion_r2115427543))
- [ ] Use upstream releases for macOS/Windows ([comment](https://github.com/docker/model-runner/pull/879#issuecomment-2829134855))


pkg/inference/backends/llamacpp/llamacpp.go
```diff
@@ -77,6 +78,27 @@ func New(
 	}, nil
 }
 
+// resolveLlamaServerBin returns the llama-server binary name to use.
+// It prefers the upstream name (llama-server) shipped by the official
+// ghcr.io/ggml-org/llama.cpp images used on Linux.  When that binary
+// does not exist in dir it falls back to the Docker-convention name
+// (com.docker.llama-server) used by macOS and Docker Desktop builds.
+func resolveLlamaServerBin(dir string) string {
```
ericcurtin, line 86: Nice

cmd/cli/pkg/standalone/controller_image.go
```diff
@@ -38,10 +38,10 @@ func controllerImageVariant(detectedGPU gpupkg.GPUSupport, backend string) strin
 		return "cuda"
 	case gpupkg.GPUSupportROCm:
 		return "rocm"
-	case gpupkg.GPUSupportMUSA:
-		return "musa"
-	case gpupkg.GPUSupportCANN:
-		return "cann"
+	case gpupkg.GPUSupportMUSA, gpupkg.GPUSupportCANN:
```
ericcurtin, line 41: Musa is available btw:

server-rocm-b8895 "2026-04-23T06:19:54.284808434Z"
server-cuda-b8895 "2026-04-23T06:33:03.536404803Z"
server-cuda12-b8895 "2026-04-23T06:33:03.536404803Z"
server-vulkan-b8895 "2026-04-23T05:27:20.537711822Z"
server-musa-b8895 "2026-04-23T05:43:19.123167181Z"
server-intel-b8895 "2026-04-23T05:33:58.173828869Z"
server-openvino-b8895 "2026-04-23T05:20:15.646954065Z"
server-cuda13-b8895 "2026-04-23T06:26:18.288770943Z"
server-b8895 "2026-04-23T05:22:57.043953717Z"

ericcurtin: By the way what makes sense to me is these:

server-rocm-b8895
server-cuda13-b8895
server-vulkan-b8895
server-musa-b8895
server-openvino-b8895

We don't need "intel", "intel" is legacy at this point. We should add "openvino" for intel though.

ericcurtin:
We can avoid building for Windows and macOS also, the releases are here:

https://github.com/ggml-org/llama.cpp/releases

ericcurtin:
It would be better if the final images became full "FROM ggml" images. Copying around binaries means risk, incompatible OS version, dependant libraries that don't get copied, etc. Full "FROM ggml" resolves many of these issues.

ericcurtin (approved):

PR #881, opened 2026-04-22
bump model-runner in pinata during release

ilopezluna:
Add steps to the bump-pinata job to update the github.com/docker/model-runner Go module in pinata (go get + go mod tidy + go mod vendor).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -477,6 +477,12 @@ jobs:
           token: ${{ secrets.CLI_RELEASE_PAT }}
           fetch-depth: 0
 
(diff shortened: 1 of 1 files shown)


doringeman (approved):
What if we also add an agent skill to call to ensure the inference manager (dmr's main.go in pinata) is corresponding with the newly updated main.go from docker/model-runner?

ilopezluna:
> What if we also add an agent skill to call to ensure the inference manager (dmr's main.go in pinata) is corresponding with the newly updated main.go from docker/model-runner?

I'll add it in a follow up PR thanks for the suggestion!

PR #892, opened 2026-04-28
add model-cli config command with INI file support

ericcurtin:
Introduce 'model-cli config' as a new top-level command with an interface
and file format inspired by, but not referencing, 'git config'.

- New cmd/cli/iniconfig package: parses and writes INI-style config files
  (section headers, subsections, boolean keys, inline comments, backslash
  escapes, quoted values, UTF-8 BOM). Writes are atomic via .lock + rename.
- New 'config' command with subcommands: get, set, unset, list, edit.
  All subcommands accept --global (default per XDG_CONFIG_HOME or
  ~/.config/model-runner/config), --system (/etc/model-runner/config),
  and --file/-f flags.
- Remove the 'config' alias from 'configure' to avoid a name collision;
  'configure' remains hidden and undocumented for existing callers.
- 'config' requires no running model-runner instance (pure local file I/O)
  and is registered outside the withStandaloneRunner group.
- Parser: handle trailing comments on section headers ([core] # comment),
  raise a clear error on lines exceeding 1 MiB, preserve existing file
  permissions on write (default 0600 for new files).
- Editor: split VISUAL/EDITOR on whitespace to support values like
  'code --wait'.
- Regenerate CLI reference docs.

### DIFF
--- a/cmd/cli/commands/config.go
+++ b/cmd/cli/commands/config.go
@@ -1,0 +1,351 @@
+package commands
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"github.com/docker/model-runner/cmd/cli/iniconfig"
+	"github.com/spf13/cobra"
+)
+
+// defaultConfigPath returns the default (global/user-level) config file path.
+// It honours XDG_CONFIG_HOME when set:
+//
+//	$XDG_CONFIG_HOME/model-runner/config
+//	~/.config/model-runner/config  (fallback)
+func defaultConfigPath() string {
+	if xdg := os.Getenv("XDG_CONFIG_HOME"); xdg != "" {
+		return filepath.Join(xdg, "model-runner", "config")
+	}
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return filepath.Join(".config", "model-runner", "config")
+	}
+	return filepath.Join(home, ".config", "model-runner", "config")
+}
+
+// systemConfigPath returns the system-wide config file path.
+func systemConfigPath() string {
+	if runtime.GOOS == "windows" {
+		if pd := os.Getenv("ProgramData"); pd != "" {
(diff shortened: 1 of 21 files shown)


cmd/cli/commands/config.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+	if system {
+		count++
+	}
+	if file != "" {
+		count++
+	}
+	if count > 1 {
+		return "", fmt.Errorf("only one of --global, --system, or --file may be specified")
+	}
```
sathiraumesh, line 57: can do something like 
```
count := 0
for _, configFlag := range []bool{global, system, file != ""} {
        if  configFlag{
            count++
        }
    }
    if count > 1 {
        return "", fmt.Errorf("only one of --global, --system, or --file may be specified")
    }

```

cmd/cli/commands/config.go
```diff
@@ -0,0 +1,351 @@
...
+					if showOrigin {
+						cmd.Printf("file:%s\t%s\n", path, v)
+					} else {
+						cmd.Println(v)
+					}
+				}
+				return nil
+			}
+
+			v, ok := f.Get(key)
```
sathiraumesh, line 171: can remove duplicate code and logic by using a slice to hold the values 
   ```
 key := args[0]

    var vals []string
    if showAll {
        vals = f.GetAll(key)
    } else if v, ok := f.Get(key); ok {
        vals = []string{v}
    }

    if len(vals) == 0 {
        if hasDefault {
            cmd.Println(defaultVal)
            return nil
        }
        return fmt.Errorf("key not found: %s", key)
    }

    for _, v := range vals {
        if showOrigin {
            cmd.Printf("file:%s\t%s\n", path, v)
        } else {
            cmd.Println(v)
        }
    }
    return nil
},
```

ilopezluna (approved):

sathiraumesh (changes requested):

PR #898, opened 2026-04-30
fix: use path.Join instead of filepath.Join for tar entry names

### LINKED ISSUE #894: Unable to get `docker model package` to package with `--chat-template`
I'm _trying_ to create a new model which has thinking disabled. My original issue, which I assumed a packaged model would resolve, is that using Open WebUI to run Qwen 3.5 9B on a remote GPU workstation results in it "thinking", no matter what. The Open WebUI option to disable thinking doesn't have any effect on the model.

Just as with the context size setting in Open WebUI not having any effect on the model's context, and since I'd had success extending the context size to 40960 using `docker model package`, my thought was that I could do the same and include a new chat template which specifically disables thinking.

Using Docker's native model called qwen3.5:9B-UD-Q4_K_XL, I've tried the following commands, which all result in the following error:

```
Failed to package model
package model: failed to load packaged model: loading model archive: load failed with status 500 Internal Server Error: error while loading model: write manifest: missing blob "sha256:a427b36b6993cda8fd07037a8ac96b06be12be49dedecd01e0c6cff13276ade8" for manifest - refusing to write unless all blobs exist
```

Note that adding or removing single or double-quotes has absolutely no effect on the outcome, so I've not included those examples. Line Feeds I've also tried CRLF (Windows) and LF (Unix) in the chat template, which I've tried as both .txt and .jinja extensions (not that it should make any difference):

### Package with from and a chat template - doesn't work.
`docker model package --from ai/qwen3.5:9B-UD-Q4_K_XL --chat-template D:\Docker\qwen3-nothinking.txt --context-size 40960 lwsrbrts/qwen35-instruct:9b-40k`

I also resorted to downloading the Unsloth GGUF from Hugging Face, the mmproj, and still using my own chat template (adding nother more than `{%- set enable_thinking = false %}` at line 3), but I get the exact same issue about a missing blob. (Side note, Unsloth are adamant that `thinking` is disabled in the 9B model, but no matter what, it still thinks when accessed from Open WebUI)

### Package from gguf, mmproj and a chat template - doesn't work.
`docker model package --gguf D:\Docker\Qwen3.5-9B-UD-Q4_K_XL.gguf --chat-template D:\Docker\qwen3-nothinking.jinja --mmproj D:\Docker\mmproj-BF16.gguf --context-size 40960 lwsrbrts/qwen35-instruct:9b-40k`

The GGUF package gives this output. All looks absolutely fine right up until the end (note that the `Transferred` info is updated when loading the GGUF and it does count up to 5.6GB, then it does the mmproj, which is 879MB):

```
PS D:\Docker> docker model package --gguf D:\Docker\Qwen3.5-9B-UD-Q4_K_XL.gguf --chat-template D:\Docker\qwen3-nothinking.jinja --mmproj D:\Docker\mmproj-BF16.gguf --context-size 40960 lwsrbrts/qwen35-instruct:9b-40k
Adding GGUF file from "D:\\Docker\\Qwen3.5-9B-UD-Q4_K_XL.gguf"
Setting context size 40960
Adding chat template file from "D:\\Docker\\qwen3-nothinking.jinja"
Adding multimodal projector file from "D:\\Docker\\mmproj-BF16.gguf"
Loading model to Model Runner...
Transferred: 879.01 MB
Failed to package model
package model: failed to load packaged model: loading model archive: load failed with status 500 Internal Server Error: error while loading model: write manifest: missing blob "sha256:a427b36b6993cda8fd07037a8ac96b06be12be49dedecd01e0c6cff13276ade8" for manifest - refusing to write unless all blobs exist
```

As mentioned, simply changing the context-size works as expected to create the model variant so all eyes are on the chat template but I have no indication what is wrong with the chat template. It is clearly not "missing" as the error suggests, nor is there anything wrong with the line feeds, name of the file, its location or other things I can think of.

ilopezluna:
## Problem

On Windows, `docker model package --from <model> --chat-template <file>` fails with:

```
missing blob sha256:... for manifest
```

## Root Cause

`tarball/target.go` used `filepath.Join` to build tar entry names. On Windows, this produces backslash-separated paths (e.g., `blobs\sha256\hex`). When the daemon (Linux) reads the tarball, the reader splits entry names on `/` and skips entries with backslashes, so all blobs are silently ignored. The manifest then references blobs that were never stored → "missing blob".

This only affects code paths that go through the tarball (i.e., when `canUseDaemonRepackage` is false): `--chat-template`, `--license`, `--mmproj`, `--gguf`, `--safetensors-dir`, `--dduf`, etc. The simple `--from --context-size` case uses a direct daemon API and was never affected.

## Fix

Replace `filepath.Join` with `path.Join` (package `path`, not `path/filepath`) in `tarball/target.go`. `path.Join` always uses forward slashes regardless of the OS.

## Test

Added `TestTargetEntryNamesUseForwardSlashes` which builds a model with GGUF + chat template, writes it to a tarball, and verifies no entry names contain backslashes.

Fixes #894

### DIFF
--- a/pkg/distribution/tarball/target.go
+++ b/pkg/distribution/tarball/target.go
@@ -5,7 +5,7 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"path/filepath"
+	"path"
 
 	"github.com/docker/model-runner/pkg/distribution/internal/progress"
 	"github.com/docker/model-runner/pkg/distribution/oci"
@@ -68,7 +68,7 @@ func (t *Target) Write(ctx context.Context, mdl types.ModelArtifact, progressWri
 		return err
 	}
 	if err = tw.WriteHeader(&tar.Header{
-		Name: filepath.Join("blobs", cn.Algorithm, cn.Hex),
+		Name: path.Join("blobs", cn.Algorithm, cn.Hex),
 		Mode: 0666,
 		Size: int64(len(rcf)),
 	}); err != nil {
@@ -97,7 +97,7 @@ func (t *Target) addLayer(layer oci.Layer, tw *tar.Writer, progressWriter io.Wri
 	if err != nil {
 		return fmt.Errorf("get layer diffID: %w", err)
 	}
-	if err := t.ensureDir(filepath.Join("blobs", diffID.Algorithm), tw); err != nil {
+	if err := t.ensureDir(path.Join("blobs", diffID.Algorithm), tw); err != nil {
 		return err
 	}
 	sz, err := layer.Size()
@@ -105,7 +105,7 @@ func (t *Target) addLayer(layer oci.Layer, tw *tar.Writer, progressWriter io.Wri
 		return fmt.Errorf("get layer size: %w", err)
 	}
 	if err = tw.WriteHeader(&tar.Header{
-		Name: filepath.Join("blobs", diffID.Algorithm, diffID.Hex),
+		Name: path.Join("blobs", diffID.Algorithm, diffID.Hex),
 		Mode: 0666,
 		Size: sz,
 	}); err != nil {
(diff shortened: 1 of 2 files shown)


ericcurtin (approved):
Looks great, thanks @ilopezluna

PR #903, opened 2026-05-06
add modelfile support to docker model package

### LINKED ISSUE #613: Add support for Modelfile
Ollama has the `Modelfile` format ([link](https://docs.ollama.com/modelfile)) inspired from the `Dockerfile` format.

I wonder if we could use the packaging command with such a file, instead of providing all the options on the command line?

areebahmeddd:
## changes

completes #628

* added `--file / -f` flag to `docker model package` to load and apply a [modelfile](https://docs.ollama.com/modelfile)

closes #613

### DIFF
--- a/cmd/cli/commands/modelfile.go
+++ b/cmd/cli/commands/modelfile.go
@@ -19,7 +19,7 @@ var modelfileAliases = map[string]string{
 	"CONTEXT-SIZE":    "CONTEXT",
 }
 
-// modelfilePathInstructions is the set of instructions whose value is a file or directory path.
+// modelfilePathInstructions is the set of instructions that take a file or directory path.
 var modelfilePathInstructions = map[string]struct{}{
 	"GGUF":            {},
 	"SAFETENSORS_DIR": {},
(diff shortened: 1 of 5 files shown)


areebahmeddd:
the command is `docker model package --file ./Modelfile`, not a new `docker model create` subcommand to match the `docker build -f Dockerfile` convention

ericcurtin:
@areebahmeddd thanks and nice work!

Two things, I think this should be "docker model build" or "docker model create", it's what people are used to from docker and ollama and it makes things simpler. Also, Ollama-type models are generally out of scope for this projects as they are not standards compliant (although we do have an Ollama-compatible API). I would like to see if there are plans to define a standard with the ModelPack folk before we proceed:

https://github.com/modelpack/model-spec/issues/208

areebahmeddd:
> Two things, I think this should be "docker model build" or "docker model create"

sure, i’ll update it post-discussion in the modelpack issue 🙂

>  Also, Ollama-type models are generally out of scope for this projects as they are not standards compliant (although we do have an Ollama-compatible API)

interesting, i just read about modelpack and its spec. ollama’s modelfile seems to not follow since it uses its own binary blob format. i agree to wait for a definitive plan from the modelpack team

ericcurtin:
> > Two things, I think this should be "docker model build" or "docker model create"
> 
> sure, i’ll update it post-discussion in the modelpack issue 🙂
> 
> > Also, Ollama-type models are generally out of scope for this projects as they are not standards compliant (although we do have an Ollama-compatible API)
> 
> interesting, i just read about modelpack and its spec. ollama’s modelfile seems to not follow since it uses its own binary blob format. i agree to wait for a definitive plan from the modelpack team

Ollama's format is not a good idea, it's a forked OCI implementation that is only compatible with ollama.com . Ollama fork lots of things in an attempt at lock-in. That's why we go with things like CNCF Modelpack, so we can be OCI registry agnostic.

areebahmeddd:
Sure, I’ll go ahead and close this PR since it’s no longer in scope

PR #911, opened 2026-05-12
fix gosec g602 error for make validate-all

### LINKED ISSUE #906: false lint issue blocking checks
i ran `make validate-all` and noticed a failure caused by a gosec g602 warning. the warning looks like a false positive since the loop already performs a bounds check but gosec does not seem to follow the control flow correctly ( [ref ](https://github.com/docker/model-runner/blob/main/pkg/logging/bracket_writer_test.go#L56))

Image

areebahmeddd:
closes #906

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,8 @@
 # Project variables
 include .versions
 
+export PATH := $(PATH):$(shell go env GOPATH)/bin
+
 APP_NAME := model-runner
 LLAMA_SERVER_VARIANT := cpu
 # Resolved lazily — only evaluated when a Docker target references it.
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,6 @@
 # Project variables
 include .versions
 
-export PATH := $(PATH):$(firstword $(subst :, ,$(shell go env GOPATH)))/bin
-
 APP_NAME := model-runner
 LLAMA_SERVER_VARIANT := cpu
 # Resolved lazily — only evaluated when a Docker target references it.
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 # Project variables
 include .versions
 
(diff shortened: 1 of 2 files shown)


pkg/logging/bracket_writer_test.go
```diff
@@ -53,14 +53,12 @@ func TestBracketWriter_MultipleLines(t *testing.T) {
...
+	for i, want := range expected {
+		matches := timestampRe.FindStringSubmatch(lines[i])
 		if len(matches) != 3 {
-			t.Errorf("line %d did not match format: %q", i, line)
-			continue
-		}
-		if matches[2] != expected[i] {
-			t.Errorf("line %d: expected %q, got %q", i, expected[i], matches[2])
+			t.Errorf("line %d did not match format: %q", i, lines[i])
+		} else if matches[2] != want {
```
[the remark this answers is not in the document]
areebahmeddd: not applicable, t.Fatalf above already prevents this

Makefile
```diff
@@ -1,6 +1,8 @@
 # Project variables
 include .versions
 
+export PATH := $(PATH):$(firstword $(subst :, ,$(shell go env GOPATH)))/bin
```
sathiraumesh: I don't think this should be done here since probably the user should set their own $PATH configs and besides if the user decides some sort other tooling like `hombrew `  this path is invalid.

sathiraumesh (changes requested):

Makefile
areebahmeddd: actually i agree, i think it's a bad practise. reverted.

ericcurtin (approved):

sathiraumesh (approved):

doringeman (approved):

PR #914, opened 2026-05-13
ci: declare contents: read on ci, e2e-test, integration-test workflows

arpitjain099:
Three CI workflows currently inherit the repo-default `GITHUB_TOKEN` scope:

- `ci.yml` (lint + tests matrix across linux/darwin/windows)
- `e2e-test.yml` (end-to-end run)
- `integration-test.yml` (integration run)

All three only do `actions/checkout` + setup + Go build/test. No git push, no GitHub API write. `contents: read` at the workflow level is the right floor.

`cli-validate.yml` in this repo already declares workflow-level `permissions: contents: read`; this brings the remaining three in line.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,9 @@ on:
   push:
     branches: [ main ]
 
+permissions:
+  contents: read
+
 jobs:
   lint:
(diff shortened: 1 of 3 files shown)


ilopezluna (approved):
Thank you!

PR #917, opened 2026-05-15
feat: add vLLM ROCm Docker variant for AMD GPUs

### LINKED ISSUE #412: vLLM ROCm
Might be possible to use vllm from:

https://hub.docker.com/r/rocm/vllm

dmedovich:
Fixes #412

Adds a `final-vllm-rocm` Docker build target that runs vLLM on AMD GPUs using upstream vLLM (no fork, no third-party wheels), per the direction in the issue.

## Approach

Builds upstream vLLM from source on top of [`rocm/vllm-dev:base`](https://hub.docker.com/r/rocm/vllm-dev) — the same base image vLLM's own `docker/Dockerfile.rocm` uses. That base already ships PyTorch ROCm, Triton ROCm, flash-attention ROCm, and the ROCm SDK pre-compiled by AMD. We just clone `vllm@v${VLLM_VERSION}` and build the wheel.

The runtime stage mirrors the `/opt/vllm-env` layout that the existing vLLM backend in [pkg/inference/backends/vllm/vllm.go](pkg/inference/backends/vllm/vllm.go) expects (via symlinks, since `rocm/vllm-dev:base` installs Python deps system-wide).

## New targets

- `make docker-build-vllm-rocm` — build the image
- `make docker-run-vllm-rocm` — build + run

`LLAMA_SERVER_VARIANT=rocm` is passed so the existing logic restricts `DOCKER_BUILD_PLATFORMS` to `linux/amd64`.

## Trade-offs to flag

1. **The ROCm image does not include llama.cpp.** The CUDA `vllm` stage builds on top of `llamacpp` and inherits the llama-server binary. For ROCm, `rocm/vllm-dev:base` and the llama.cpp ROCm image use incompatible ROCm runtime / PyTorch versions, so `vllm-rocm` is a vLLM-only artifact. Happy to revisit if you'd prefer a combined shape.
2. **Build is heavy.** ~30–60 min (HIP kernel compilation), ~12–15 GB final image — likely needs its own CI treatment.
3. **`PYTORCH_ROCM_ARCH` defaults to `gfx90a;gfx942;gfx1100;gfx1101`** (MI200/MI300 + RDNA3 7900/7800) — what upstream vLLM officially supports. Exposed as a Dockerfile ARG, overridable at build time.

## ttested

- ✅ `go build ./...`, `go test ./...`, `go vet`, `golangci-lint` — clean
- ✅ `make help` shows the new targets
- ❌ Did **not** run `make docker-build-vllm-rocm` end-to-end. I don't have a supported GPU (only an RX 6600 = gfx1032, not in vLLM's official ROCm targets), so a runtime test wouldn't be informative even if the build succeeded.

Would appreciate a sanity check on the architectural choice (vLLM-only ROCm image vs combined). Happy to adjust.

## History

Original commit (`76db8d3`, force-pushed) installed vLLM via pip from `wheels.vllm.ai/${VLLM_VERSION}/rocm63`, but that URL doesn't exist — vLLM only publishes CUDA wheels there. Rewrote to source-build approach above.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -155,3 +155,66 @@ COPY --from=builder /app/model-runner /app/model-runner
 FROM sglang AS final-sglang
 # Copy the built binary from builder-sglang (without vLLM)
 COPY --from=builder-sglang /app/model-runner /app/model-runner
+
+# --- vLLM ROCm: builder stage ---
+# Builds upstream vLLM from source on AMD's pre-built ROCm dev image, which
+# already contains PyTorch ROCm, Triton, flash-attention, and the ROCm SDK
+# (see https://hub.docker.com/r/rocm/vllm-dev). vLLM is checked out at the
+# tagged release matching VLLM_VERSION — no fork, no custom wheels.
+FROM rocm/vllm-dev:base AS vllm-rocm-builder
+
+ARG VLLM_VERSION=0.19.1
+# Target GPU architectures officially supported by vLLM ROCm:
+# gfx90a (MI200), gfx942 (MI300), gfx1100/1101 (RDNA3 7900/7800).
+ARG PYTORCH_ROCM_ARCH="gfx90a;gfx942;gfx1100;gfx1101"
+ENV PYTORCH_ROCM_ARCH=${PYTORCH_ROCM_ARCH}
+
+RUN git clone --depth 1 --branch v${VLLM_VERSION} \
+    https://github.com/vllm-project/vllm.git /vllm-src
+
+WORKDIR /vllm-src
+RUN python3 -m pip install --no-cache-dir -r requirements/rocm.txt \
+    && python3 setup.py bdist_wheel --dist-dir=/wheels
+
+# --- vLLM ROCm: runtime stage ---
+# Mirrors the /opt/vllm-env layout that pkg/inference/backends/vllm/vllm.go
+# expects (binary at /opt/vllm-env/bin/vllm, version file at
+# /opt/vllm-env/version). Symlinks are used instead of a real venv because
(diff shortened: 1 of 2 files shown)


dmedovich:
@ericcurtin yo! look it when you have a free time. 

Thanks!

ericcurtin (approved):

ericcurtin:
Very nice @dmedovich ... You could do the project a huge favour actually... We are lacking people with AMD GPUs, if you could confirm llama.cpp/vLLM works with AMD GPUs that would be great! We are also supposed to automatically detect when your machine has an AMD GPU and just do the right thing in that case, that's another thing, we don't know if it works...

Merging on green build

dmedovich:
@ericcurtin 
Checked the auto-detection code — it calls docker.Info() via the Moby client 
to look up registered runtimes (nvidia/rocm/musa). I'm running Podman, not 
Docker, so I can't test that path rn. Worth noting that Podman users would always 
get GPUSupportNone regardless of what GPU they have.

Will install Docker a bit later and check all three points, will come back with detailed info on each.

ericcurtin:
@dmedovich also vllm suggests this to install for ROCm (https://vllm.ai/):

uv pip install vllm --extra-index-url https://wheels.vllm.ai/rocm/0.21.0/rocm722

why don't we go with that? (would probably reduce build time too)

dmedovich:
@ericcurtin  yep, this will indeed simplify and speed up the build, while also preserving the base version of llamacpp ROCm so that llama.cpp itself remains available as a backup (like the cuda example). look here #923

PR #922, opened 2026-05-19
Add dmr sandbox config and launch command

### LINKED ISSUE #915: Docker Sandboxes and DMR integration
It should be possible to do this:

```
dmr config sandbox.tool sbx
dmr launch opencode
```

`dmr config` should be git-style
`dmr launch` should be ollama-style

config format should be toml-style

CoderHariswar:
Fixes #915

#### Describe the changes you have made in this PR -

This PR adds support for configuring and launching tools through a sandbox from the `dmr` wrapper.

Users can now configure the sandbox tool with:

```bash
dmr config sandbox.tool sbx
```

The above command writes TOML-style config

Manual Testing done using the following commands 

```
tmpdir="$(mktemp -d)"
XDG_CONFIG_HOME="$tmpdir/config" ./dmr config sandbox.tool sbx
cat "$tmpdir/config/dmr/config.toml"
```

Tested launch through a fake sandbox command and outputs have been verified successfully

### DIFF
--- a/cmd/cli/commands/configure.go
+++ b/cmd/cli/commands/configure.go
@@ -11,10 +11,9 @@ func newConfigureCmd() *cobra.Command {
 	var flags ConfigureFlags
 
 	c := &cobra.Command{
-		Use:     "configure [--context-size=<n>] [--speculative-draft-model=<model>] [--hf_overrides=<json>] [--gpu-memory-utilization=<float>] [--mode=<mode>] [--think] [--keep-alive=<duration>] MODEL [-- <runtime-flags...>]",
-		Aliases: []string{"config"},
-		Short:   "Manage model runtime configurations",
-		Hidden:  true,
+		Use:    "configure [--context-size=<n>] [--speculative-draft-model=<model>] [--hf_overrides=<json>] [--gpu-memory-utilization=<float>] [--mode=<mode>] [--think] [--keep-alive=<duration>] MODEL [-- <runtime-flags...>]",
+		Short:  "Manage model runtime configurations",
+		Hidden: true,
 		Args: func(cmd *cobra.Command, args []string) error {
 			argsBeforeDash := cmd.ArgsLenAtDash()
 			if argsBeforeDash == -1 {
--- a/cmd/cli/commands/launch.go
+++ b/cmd/cli/commands/launch.go
@@ -196,30 +196,28 @@ Examples:
 }
 
 func launchSandboxedHostApp(cmd *cobra.Command, sandboxTool, app string, appArgs []string, dryRun bool) error {
-	if strings.ContainsAny(sandboxTool, "\x00\r\n") {
-		return fmt.Errorf("sandbox.tool contains invalid characters")
-	}
-
-	sandboxToolPath, err := exec.LookPath(sandboxTool)
-	if err != nil {
-		return fmt.Errorf("sandbox tool %q not found in PATH: %w", sandboxTool, err)
+	if err := validateSandboxTool(sandboxTool); err != nil {
+		return err
 	}
 
 	args := append([]string{app}, appArgs...)
 
-	if dryRun {
-		cmd.Printf("%s %s\n", sandboxToolPath, strings.Join(args, " "))
-		return nil
+	switch sandboxTool {
+	case "sbx":
+		if dryRun {
+			cmd.Printf("sbx %s\n", strings.Join(args, " "))
+			return nil
+		}
+
+		launchCmd := exec.Command("sbx", args...)
+		launchCmd.Stdin = os.Stdin
+		launchCmd.Stdout = os.Stdout
+		launchCmd.Stderr = os.Stderr
+
+		return launchCmd.Run()
+	default:
+		return fmt.Errorf("unsupported sandbox tool %q", sandboxTool)
 	}
-
-	// #nosec G204 -- sandboxToolPath is an explicit user-configured executable
-	// selected from allowedSandboxTools. exec.Command does not invoke a shell.
-	launchCmd := exec.Command(sandboxToolPath, args...)
-	launchCmd.Stdin = os.Stdin
-	launchCmd.Stdout = os.Stdout
-	launchCmd.Stderr = os.Stderr
-
-	return launchCmd.Run()
 }
 
 // listSupportedApps prints all supported apps with their descriptions and install status.
--- a/cmd/cli/commands/launch.go
+++ b/cmd/cli/commands/launch.go
@@ -147,15 +147,12 @@ Examples:
 			// resolving runner endpoints. This keeps sandbox launch independent of
 			// whether the host app binary itself is installed.
 			if _, ok := hostApps[app]; ok {
-				sandboxTool, err := readSandboxToolConfig()
+				sandboxTool, err := configuredSandboxTool()
 				if err != nil {
 					return err
 				}
 
 				if sandboxTool != "" {
-					if err := validateSandboxTool(sandboxTool); err != nil {
-						return err
-					}
 					return launchSandboxedHostApp(cmd, sandboxTool, app, appArgs, dryRun)
 				}
 			}
@@ -178,7 +175,6 @@ Examples:
 			if ca, ok := containerApps[app]; ok {
 				return launchContainerApp(cmd, ca, ep.container, image, port, detach, appArgs, dryRun)
 			}
-
 			if cli, ok := hostApps[app]; ok {
 				return launchHostApp(cmd, app, ep.host, cli, model, runner, appArgs, dryRun)
 			}
@@ -196,28 +192,9 @@ Examples:
 }
 
 func launchSandboxedHostApp(cmd *cobra.Command, sandboxTool, app string, appArgs []string, dryRun bool) error {
-	if err := validateSandboxTool(sandboxTool); err != nil {
-		return err
-	}
-
 	args := append([]string{app}, appArgs...)
 
-	switch sandboxTool {
-	case "sbx":
-		if dryRun {
-			cmd.Printf("sbx %s\n", strings.Join(args, " "))
-			return nil
-		}
-
-		launchCmd := exec.Command("sbx", args...)
-		launchCmd.Stdin = os.Stdin
-		launchCmd.Stdout = os.Stdout
-		launchCmd.Stderr = os.Stderr
-
-		return launchCmd.Run()
-	default:
-		return fmt.Errorf("unsupported sandbox tool %q", sandboxTool)
-	}
+	return runSandboxTool(cmd, sandboxTool, args, dryRun)
 }
 
 // listSupportedApps prints all supported apps with their descriptions and install status.
--- a/cmd/cli/commands/launch.go
+++ b/cmd/cli/commands/launch.go
@@ -166,12 +166,10 @@ Examples:
 			if err != nil {
 				return err
 			}
-
 			// --config: print configuration without launching
 			if configOnly {
 				return printAppConfig(cmd, app, ep, image, port)
 			}
-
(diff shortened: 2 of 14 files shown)


cmd/dmr/main.go
```diff
@@ -67,3 +80,121 @@ func newServeCmd() *cobra.Command {
...
+		return fmt.Errorf("sandbox.tool is not configured. Run: dmr config sandbox.tool <tool>")
+	}
+
+	cmd := exec.Command(sandboxTool, args...)
+	cmd.Stdin = os.Stdin
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	return cmd.Run()
+}
```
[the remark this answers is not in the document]
CoderHariswar: @copilot apply changes based on [this feedback](https://github.com/docker/model-runner/pull/922#discussion_r3266167545)

cmd/dmr/main.go
```diff
@@ -27,6 +31,15 @@ func main() {
...
 
 func run() error {
+	if len(os.Args) > 1 {
+		switch os.Args[1] {
+		case "config":
+			return runConfig(os.Args[2:])
+		case "launch":
+			return runLaunch(os.Args[2:])
+		}
+	}
```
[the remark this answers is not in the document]
CoderHariswar: @copilot apply changes based on [this feedback](https://github.com/docker/model-runner/pull/922#discussion_r3266167601)

cmd/dmr/main.go
```diff
@@ -67,3 +80,121 @@ func newServeCmd() *cobra.Command {
...
+		key, value, ok := strings.Cut(line, "=")
+		if !ok {
+			continue
+		}
+
+		if strings.TrimSpace(key) != "tool" {
+			continue
+		}
+
+		return strings.Trim(strings.TrimSpace(value), `"`), nil
```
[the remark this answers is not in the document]
CoderHariswar: @copilot apply changes based on [this feedback](https://github.com/docker/model-runner/pull/922#discussion_r3266167637)

cmd/dmr/main.go
```diff
@@ -67,3 +80,124 @@ func newServeCmd() *cobra.Command {
...
+		return fmt.Errorf("usage: dmr config sandbox.tool <tool>")
+	}
+
+	key := args[0]
+	value := args[1]
+
+	if key != "sandbox.tool" {
+		return fmt.Errorf("unsupported config key %q", key)
+	}
+
```
sathiraumesh: perhaps much better if we can add a list of supported tools in the top of the file and only support here like
```

var allowedSandboxTools = map[string]struct{}{
    "firejail":   {},
    "bubblewrap": {},
    "bwrap":      {},
}
```
```

if _, ok := allowedSandboxTools[value]; !ok {
        return fmt.Errorf("unsupported sandbox tool %q", sandboxTool)
    }
```

cmd/dmr/main.go
```diff
@@ -67,3 +80,124 @@ func newServeCmd() *cobra.Command {
...
+func runLaunch(args []string) error {
+	if len(args) == 0 {
+		return fmt.Errorf("usage: dmr launch <tool> [args...]")
+	}
+
+	sandboxTool, err := readSandboxToolConfig()
+	if err != nil {
+		return err
+	}
+
```
sathiraumesh: same thing like above can be done here for verifying supported tools

sathiraumesh (changes requested):

sathiraumesh:
@CoderHariswar something i noticed when reviewing is. Is there any reason you didn't use the same pattern in the implementation like registering a subcommand like the `serveCmd` here in this file.

cmd/dmr/main.go
ericcurtin: Sure but lets just start with sbx for this PR, these can be follow on PRs

ericcurtin:
We must ensure this works for model-cli and dmr. And test, test, test to see this works as expected

cmd/dmr/main.go
sathiraumesh: @ericcurtin yes, this was my exact point, like it's better if we can specify the supported ones, both here and the launch

CoderHariswar:
Thanks for the review @sathiraumesh 

I was confused at first because `config` was already an alias for `configure`, so `dmr config sandbox.tool sbx` was getting routed to the existing configure command.

My first fix handled `config` directly inside `cmd/dmr`, but after I got stuck and asked GPT for help, I understood that this was not the right pattern here. @ericcurtin also mentioned that this should work for both model-cli and dmr.

I updated the PR to follow that approach:
- moved the config support into the shared `cmd/cli/commands` package
- registered it as a normal Cobra subcommand
- kept `dmr` using `commands.NewRootCmd`
- removed the old `config` alias from `configure`
- restricted supported sandbox tools to `sbx` for this PR
- added validation during config and launch
- updated tests for config and launch behavior

Tested locally with:

go test ./cmd/cli/commands
go test ./cmd/dmr
go build -o dmr ./cmd/dmr

I also manually tested `dmr config sandbox.tool sbx` and `dmr launch opencode` with a fake `sbx` script.

ericcurtin:
> Thanks for the review @sathiraumesh
> 
> I was confused at first because `config` was already an alias for `configure`, so `dmr config sandbox.tool sbx` was getting routed to the existing configure command.
> 
> My first fix handled `config` directly inside `cmd/dmr`, but after I got stuck and asked GPT for help, I understood that this was not the right pattern here. @ericcurtin also mentioned that this should work for both model-cli and dmr.
> 
> I updated the PR to follow that approach:
> 
> * moved the config support into the shared `cmd/cli/commands` package
> * registered it as a normal Cobra subcommand
> * kept `dmr` using `commands.NewRootCmd`
> * removed the old `config` alias from `configure`
> * restricted supported sandbox tools to `sbx` for this PR
> * added validation during config and launch
> * updated tests for config and launch behavior
> 
> Tested locally with:
> 
> go test ./cmd/cli/commands go test ./cmd/dmr go build -o dmr ./cmd/dmr
> 
> I also manually tested `dmr config sandbox.tool sbx` and `dmr launch opencode` with a fake `sbx` script.

We want to do a brand new "config", "git config"-style, using toml, the old "configure" we should deprecate it's kinda strange, it only lives as long as the daemon

CoderHariswar:
Got it @ericcurtin  for this PR I  focused on `sandbox.tool` with `sbx` as the only supported sandbox tool.

cmd/cli/commands/configure.go
```diff
@@ -11,10 +11,9 @@ func newConfigureCmd() *cobra.Command {
 	var flags ConfigureFlags
 
 	c := &cobra.Command{
-		Use:     "configure [--context-size=<n>] [--speculative-draft-model=<model>] [--hf_overrides=<json>] [--gpu-memory-utilization=<float>] [--mode=<mode>] [--think] [--keep-alive=<duration>] MODEL [-- <runtime-flags...>]",
-		Aliases: []string{"config"},
-		Short:   "Manage model runtime configurations",
-		Hidden:  true,
+		Use:    "configure [--context-size=<n>] [--speculative-draft-model=<model>] [--hf_overrides=<json>] [--gpu-memory-utilization=<float>] [--mode=<mode>] [--think] [--keep-alive=<duration>] MODEL [-- <runtime-flags...>]",
+		Short:  "Manage model runtime configurations",
```
sathiraumesh, line 15: I was hoping this wouldn't be removed we could keep this like this without removing if not it would break the style of `dmr config `or `docker model config ` style .

What i suggest is we register it as a sub command in config. so one can` do dmr config sandbox.tool ... or docker model config sandbox.tool`

like `c.AddCommand(newSandboxConfigCmd())` (note that sandbox.tool is not lke a cmd style format but not sure in this case there is any other approach)

cmd/cli/commands/launch.go
```diff
@@ -155,9 +172,11 @@ Examples:
 			if ca, ok := containerApps[app]; ok {
 				return launchContainerApp(cmd, ca, ep.container, image, port, detach, appArgs, dryRun)
 			}
+
```
sathiraumesh: can remove these line space changes which i think is not part of changes

cmd/cli/commands/launch.go
```diff
@@ -170,6 +189,31 @@ Examples:
 	return c
 }
 
+func launchSandboxedHostApp(cmd *cobra.Command, sandboxTool, app string, appArgs []string, dryRun bool) error {
+	if err := validateSandboxTool(sandboxTool); err != nil {
```
sathiraumesh: much better if we can retrieve so everything related to the sandbox happens inside the sandbox.go
```

err, validatedSbxTool = `validateSandboxTool(sandboxTool)

if err != nil {
 return err
}

```

Then remove the `switch ... ` and directly 

```
 if dryRun {
			cmd.Printf("sbx %s\n", strings.Join(args, " "))
			return nil
		}

		launchCmd := exec.Command("sbx", args...)
		launchCmd.Stdin = os.Stdin
		launchCmd.Stdout = os.Stdout
		launchCmd.Stderr = os.Stderr

		return launchCmd.Run()
```

cmd/cli/commands/sandbox.go
```diff
@@ -0,0 +1,126 @@
...
+			if err := validateSandboxTool(value); err != nil {
+				return err
+			}
+
+			return writeSandboxToolConfig(value)
+		},
+	}
+}
+
+func validateSandboxTool(tool string) error {
```
sathiraumesh: can change as above mentioned suggestions

cmd/cli/commands/root.go
```diff
@@ -120,6 +120,7 @@ func NewRootCmd(cli *command.DockerCli) *cobra.Command {
 		newShowCmd(),
 		newComposeCmd(),
 		newLaunchCmd(),
+		newSandboxConfigCmd(),
```
sathiraumesh: we don't need to do this if we do as a configs sub command

sathiraumesh:
Probably you might need to check if `sbx` is not configured it fail tests locally. When i tried in my computer it did

cmd/cli/commands/root.go
ericcurtin: Should be newConfigCmd, this will be a generic config system, sandbox is just one of the things

CoderHariswar:
@sathiraumesh Updated based on review

I changed `config` to a generic parent command through `newConfigCmd()`, and registered `sandbox.tool` under it. So the command now stays as:

`dmr config sandbox.tool sbx`

and the same structure works for:

`docker model config sandbox.tool sbx`

I also moved the sandbox validation/config helpers into `sandbox.go`, and `launch.go` now only checks if a sandbox tool is configured and delegates to the sandbox helper.

For this PR, `sbx` is still the only supported sandbox tool. I updated the tests and regenerated the CLI docs with `make docs`.

cmd/cli/commands/launch.go
```diff
@@ -158,6 +172,7 @@ Examples:
 			if cli, ok := hostApps[app]; ok {
 				return launchHostApp(cmd, app, ep.host, cli, model, runner, appArgs, dryRun)
 			}
+
```
sathiraumesh, line 179: remove extra change

sathiraumesh (changes requested):

sathiraumesh:
Other things seems fine the only thing im not sure is since we remove the config alias from the config command here it will not work `dmr config ...` for others. @ericcurtin this okay for now ?.We can use this as a base for the new config command implementation for the rest of configs

ericcurtin:
Build failing

ericcurtin:
Kicked off a build again

doringeman:
Hello 👋 FWIW we could also add a kit for Model Runner to work with https://docs.docker.com/ai/sandboxes/. Anyone interested in this?

PR #923, opened 2026-05-19
refactor: install vLLM ROCm from upstream wheels

dmedovich:
Replaces the build-from-source approach (introduced in #917) with the upstream vLLM ROCm wheels published at https://wheels.vllm.ai/rocm/, as suggested by @ericcurtin.

Builds now take ~3-5 min instead of ~30-60 min, and the runtime image keeps the llamacpp ROCm base — restoring llama.cpp as an in-image fallback (matching the shape of the CUDA vllm variant).

vLLM ROCm wheels follow a separate release cadence from CUDA, so VLLM_ROCM_VERSION (default 0.21.0) is tracked independently from VLLM_VERSION (CUDA, default 0.19.1).

Refs #412

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -153,64 +153,38 @@ FROM sglang AS final-sglang
 # Copy the built binary from builder-sglang (without vLLM)
 COPY --from=builder-sglang /app/model-runner /app/model-runner
 
-# --- vLLM ROCm: builder stage ---
-# Builds upstream vLLM from source on AMD's pre-built ROCm dev image, which
-# already contains PyTorch ROCm, Triton, flash-attention, and the ROCm SDK
-# (see https://hub.docker.com/r/rocm/vllm-dev). vLLM is checked out at the
-# tagged release matching VLLM_VERSION — no fork, no custom wheels.
-FROM rocm/vllm-dev:base AS vllm-rocm-builder
-
-ARG VLLM_VERSION=0.19.1
-# Target GPU architectures officially supported by vLLM ROCm:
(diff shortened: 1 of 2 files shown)


ericcurtin (approved):

ericcurtin:
@dmedovich sometimes the bots are wrong, but can you confirm if they are right or wrong?

dmedovich:
@ericcurtin happy to confirm:

**Gemini's review — wrong on all three points:**

1. *"wheels.vllm.ai/rocm is a flat repo without versioned subdirs"* — false. Verified:
   ```
   $ curl -sI https://wheels.vllm.ai/rocm/0.21.0/rocm722/
   HTTP/2 200
   ```
   The index returns `vllm-0.21.0+rocm722-cp312-cp312-manylinux_2_34_x86_64.whl`.


2. *"Needs `v` prefix (`v0.21.0`)"* — false. The existing CUDA stage in the same Dockerfile uses `wheels.vllm.ai/${VLLM_VERSION}/cu130` without a `v` prefix and works fine.

3. *"`rocm722` is a typo, should be `rocm61`/`rocm62`"* — false. `rocm722` is exactly the suffix you suggested in your comment on #917 (`https://wheels.vllm.ai/rocm/0.21.0/rocm722`), which this PR implements.

**Sourcery's review — correct, fixed in last commit (2dbc9086):**

Without pinning `vllm==${VLLM_ROCM_VERSION}`, uv resolves against both PyPI and our extra index. If a newer vLLM (say `0.22.0`) lands on PyPI before the matching ROCm wheel is published, `--index-strategy unsafe-best-match` would pick the higher PyPI version — silently installing a CUDA-only wheel into the ROCm image. Now pinned.

PR #927, opened 2026-05-21
feat: expose model configure command

Montana:
Remove the hidden flag from the configure command so users can access model runtime configuration management directly from the CLI.

### DIFF
--- a/cmd/cli/commands/configure.go
+++ b/cmd/cli/commands/configure.go
@@ -4,19 +4,19 @@ import (
 	"fmt"
 
 	"github.com/docker/model-runner/cmd/cli/commands/completion"
+
 	"github.com/spf13/cobra"
 )
 
 func newConfigureCmd() *cobra.Command {
 	var flags ConfigureFlags
-
 	c := &cobra.Command{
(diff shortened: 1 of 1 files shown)


cmd/cli/commands/configure.go
```diff
@@ -4,19 +4,19 @@ import (
...
 )
 
 func newConfigureCmd() *cobra.Command {
 	var flags ConfigureFlags
-
 	c := &cobra.Command{
 		Use:     "configure [--context-size=<n>] [--speculative-draft-model=<model>] [--hf_overrides=<json>] [--gpu-memory-utilization=<float>] [--mode=<mode>] [--think] [--keep-alive=<duration>] MODEL [-- <runtime-flags...>]",
 		Aliases: []string{"config"},
 		Short:   "Manage model runtime configurations",
-		Hidden:  true,
```
sathiraumesh, line 17: I think this is not a needed change since thins might be deprecated moving forward to have more better config style may be. Correct me if im wrong @ericcurtin

ericcurtin:
Yeah this is planned for deprecation would rather not expose

PR #936, opened 2026-05-22
Minimal SECURITY.md which is expected in github repos

schneidergithub:
A simple security policy, since I'm sure the owners of the repo will want to update this.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,5 @@
+# Security Policy
+
+## Reporting a Vulnerability
+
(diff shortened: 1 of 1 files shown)


schneidergithub:
@ericcurtin - I'm curious why this was closed?  There is no security.md file, and at a bare minimum an email address would be useful to know, so people (like me) can submit security reports responsibly. I imagine a likely scenario is someone posting publicly a vulnerability in the repo's issues.

PR #938, opened 2026-05-25
cli: hint when install-runner options are ignored

### LINKED ISSUE #907: [Bug]vllm backend doesn't work for wsl2 ubuntu22.04
## Systeminfo
Docker Desktop v4.71.0
WSL Ubuntu 22.04
Windows 11 Pro
Nvidia Geforce RTX 4090

## Reproduce issue:
1. engine status
```
docker model status
Docker Model Runner is running
BACKEND    STATUS         DETAILS                                                                                       
llama.cpp  Running        llama.cpp latest-cuda (sha256:48883a67749cbe5343eacdc666bb1348ea70a55e3702b50059e20e6468ae7be1) e365e65
diffusers  Not Installed                                                                                                
vllm       Not Installed  only supported on Linux  
```
---
2. installation command
```
docker model install-runner --backend vllm --gpu cuda
Model Runner container docker-model-runner (85475cc8896c) is already running
```
---
3. model serving failed
```
docker model run hf.co/zed-industries/zeta-2 -d
Failed to load model: preload failed: status=500 body=unable to load runner: error waiting for runner to be ready: failed to get args for llama.cpp: GGUF file required by llama.cpp backend
```
---
4. gpu accessability possible issue
```
docker run --rm --gpus all nvidia/cuda:12.0-base nvidia-smi
Unable to find image 'nvidia/cuda:12.0-base' locally
docker: Error response from daemon: manifest for nvidia/cuda:12.0-base not found: manifest unknown: manifest unknown

Run 'docker run --help' for more information
```
---
5. local gpu availability 
```
nvidia-smi
Fri May  8 07:39:55 2026
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 590.57                 Driver Version: 591.86         CUDA Version: 13.1     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA GeForce RTX 4090        On  |   00000000:01:00.0 Off |                  Off |
| 35%   31C    P8             10W /  450W |     384MiB /  24564MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|  No running processes found                                                             |
+-----------------------------------------------------------------------------------------+
```

Yash-2003P:
## Summary
-  Add a reinstall hint when `install-runner` finds an existing running Model Runner container
-  Show the hint only when the user explicitly passed `--backend` or `--gpu`
-  Preserve the existing “already running” message and return behavior
-  Add focused tests for the new hint-generation logic

Fixes #907

## What changed
-  Added `commandFlagChanged` to check whether a Cobra flag was explicitly changed
-  Added `existingRunnerOptionsHint` to build the reinstall command hint
-  Updated the existing `ctrID != ""` branch to print the hint after the current already-running message
-  Added `cmd/cli/commands/install_runner_existing_options_hint_test.go`

## What this does not change
- Does not automatically reinstall or recreate an existing runner
- Does not change `reinstall-runner` behavior
- Does not change backend selection logic
- Does not change GPU probing or CUDA/vLLM runtime behavior
- Does not attempt to fully fix WSL2 CUDA/vLLM execution

## Tests added
- No hint when `--backend` and `--gpu` were not explicitly passed
- Backend-only hint
- GPU-only CUDA hint
- Backend + CUDA hint
- Explicit `--gpu none` hint
- Defensive flag-check cases

## Validation
- `go test ./cmd/cli/commands -run "TestInstallRunner|TestExistingRunnerOptionsHint|TestCommandFlagChanged" -v`
- go test ./cmd/cli/commands -v`
- `go test ./...`
- `make validate-all`

## Validation details
- `go.mod` tidy check passed
- `golangci-lint run ./...` passed with 0 issues
- `go test -race ./...` passed
- Shellcheck validation passed
- `.versions` is in sync with Dockerfile ARGs

### DIFF
--- a/cmd/cli/commands/install-runner.go
+++ b/cmd/cli/commands/install-runner.go
@@ -245,6 +245,39 @@ type runnerOptions struct {
 	tlsKey          string
 }
 
+func commandFlagChanged(cmd *cobra.Command, name string) bool {
+	if cmd == nil {
+		return false
+	}
+	flag := cmd.Flags().Lookup(name)
+	return flag != nil && flag.Changed
+}
+
+func existingRunnerOptionsHint(cmd *cobra.Command, opts runnerOptions) string {
+	backendChanged := commandFlagChanged(cmd, "backend")
+	gpuChanged := commandFlagChanged(cmd, "gpu")
+
+	if !backendChanged && !gpuChanged {
+		return ""
+	}
+
+	reinstallArgs := []string{"docker", "model", "reinstall-runner"}
+
+	if backendChanged && opts.backend != "" {
+		reinstallArgs = append(reinstallArgs, "--backend", fmt.Sprintf("%q", opts.backend))
+	}
+	if gpuChanged && opts.gpuMode != "" {
+		reinstallArgs = append(reinstallArgs, "--gpu", fmt.Sprintf("%q", opts.gpuMode))
+	}
+
+	return fmt.Sprintf(
+		"\nThe requested runner options were not applied because the Model Runner container is already running.\n"+
+			"To recreate the runner with the requested options, run:\n\n"+
+			"  %s\n",
+		strings.Join(reinstallArgs, " "),
+	)
+}
+
 // runInstallOrStart is shared logic for install-runner and start-runner commands
 func runInstallOrStart(cmd *cobra.Command, opts runnerOptions, debug bool) error {
 	// On macOS ARM64, the vllm backend requires deferred installation
@@ -343,6 +376,9 @@ func runInstallOrStart(cmd *cobra.Command, opts runnerOptions, debug bool) error
 			} else {
 				cmd.Printf("Model Runner container %s is already running\n", ctrID[:12])
 			}
+
+			cmd.Print(existingRunnerOptionsHint(cmd, opts))
+
 			return nil
 		}
 	}
--- a/cmd/cli/commands/install-runner_test.go
+++ b/cmd/cli/commands/install-runner_test.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/docker/model-runner/pkg/inference/backends/llamacpp"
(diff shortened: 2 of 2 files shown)


cmd/cli/commands/install_runner_existing_options_hint_test.go
```diff
@@ -0,0 +1,126 @@
+package commands
```
sathiraumesh: Better if you can put the tests in existing `install-runner_test.go` file

sathiraumesh (changes requested):

cmd/cli/commands/install_runner_existing_options_hint_test.go
Yash-2003P: Ok , I do that. May you please help me with if the gemini suggested changes to be done in code ?

sathiraumesh: @YasharthPanwar-2003  if you check the code both of the` opts.backend ` and the `opts.gpuMode` are validated by predefined sepcs. so it won't be a problem though but for defense in future its better to do the suggestion by `gemini `here  like change to the below 
```

if backendChanged && opts.backend != "" {
		reinstallArgs = append(reinstallArgs, "--backend", fmt.Sprintf("%q", opts.backend))
	}
	if gpuChanged && opts.gpuMode != "" {
		reinstallArgs = append(reinstallArgs, "--gpu", fmt.Sprintf("%q", opts.gpuMode))
	}
```

Its always important to double check what these analyzers suggest because it sometimes can be wrong. Always try to check whether its the case

sathiraumesh: and test this feature in local as well :)

Yash-2003P: I dont have NVIDIA gpu cuda , but I tested it over wsl2 Ubuntu 22.04 for cpu only , after the new updation I will test again!

ericcurtin:
@YasharthPanwar-2003 thanks for looking at this, is the bot comment worth fixing?

ericcurtin:
You will need to fix CI also

Yash-2003P:
Updated the PR based on review feedback.

Summary:
- Moved the new hint tests into the existing install-runner test file.
- Quoted displayed backend/GPU values in the suggested reinstall command.
- Added a defensive test for quoted flag values.
- Updated CI and the model-cli validation workflow to avoid the matrix subaction that pulled an unpinned transitive action, while keeping the prepare job.

Manual verification:
- Verified Docker Desktop Model Runner works on WSL2 Ubuntu 22.04 with a CPU model:
  `docker model run ai/smollm2 "Say hello in one short sentence."`
- Verified the local edited CLI prints the new hint when an existing `docker-model-runner` container is already running:
  - `--gpu none` suggests `docker model reinstall-runner --gpu "none"`
  - `--backend vllm` suggests `docker model reinstall-runner --backend "vllm"`
  - `--backend vllm --gpu cuda` suggests `docker model reinstall-runner --backend "vllm" --gpu "cuda"`

Validation:
- Focused install-runner tests passed on WSL2 Ubuntu 22.04.
- Focused install-runner tests passed on Windows 11.
- Docker Bake validate-docs and validate-tests targets printed correctly.
- Full `make validate-all` passed on WSL2 Ubuntu 22.04

I do not have NVIDIA CUDA hardware, so I did not run the actual vLLM CUDA reinstall path. The manual test verifies the intended existing-runner hint behavior without requiring CUDA.

.github/workflows/cli-validate.yml
```diff
@@ -29,16 +29,11 @@ jobs:
     outputs:
       matrix: ${{ steps.generate.outputs.matrix }}
     steps:
-      -
-        name: Checkout
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
       -
         name: Generate matrix
         id: generate
-        uses: docker/bake-action/subaction/matrix@a66e1c87e2eca0503c343edf1d208c716d54b8a8
```
sathiraumesh: any reason why you change this ?

Yash-2003P: Yes. The reason was the failing `Validate model-cli / prepare` check.

So , previously before new commit , it showed as gvien below . further , I understood to do changes in CI as per asked by ericcurtin

The failure log showed:

`actions/github-script@v8 is not allowed in docker/model-runner because all actions must be pinned to a full-length commit SHA`

That check comes from this workflow (`Validate model-cli`) and specifically the `prepare` job. The direct `docker/bake-action/subaction/matrix` reference is pinned, but the subaction path was still causing GitHub to resolve `actions/github-script@v8`, which is blocked by the repo/org full-SHA action policy.

GitHub documents that repositories/organizations can require actions to be pinned to a full-length commit SHA:
https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository

Since `cmd/cli/docker-bake.hcl` currently has only two validation targets under the `validate` group (`validate-docs` and `validate-tests`), I replaced the dynamic matrix generation with an explicit matrix output while keeping the existing `prepare` and `validate` job structure.

If you prefer to keep the bake matrix subaction and handle this CI policy another way, I can revert the workflow change and keep this PR scoped to the install-runner behavior.

sathiraumesh: I think we should not do this change in the CI. I think what @ericcurtin mentioned was to fix reason if any changes done from this PR breaks the CI.

Yash-2003P: I was thinking same ealier . Thanks for feedback  , I will undo it and keep the relevant ones . 

Updated based on the review feedback.

I reverted the CI workflow change and kept this PR scoped to the install-runner behavior only.

ericcurtin:
We have conflicts here

.github/workflows/cli-validate.yml
ericcurtin: All these changes to this file should be reverted like @sathiraumesh said... I meant the build was failing in this PR, we cannot merge until it's green

sathiraumesh (approved):
LGTM !

Yash-2003P:
@ericcurtin May you please review it and also any more changes needed . Awaiting for review and feedback

ilopezluna (approved):
Thank you for your contribution @Yash-2003P 🙏

Yash-2003P:
Thanks a lot to @ericcurtin  , @sathiraumesh  for helping me out in review and @ilopezluna for merging .

PR #947, opened 2026-05-31
sync docker credentials to container

### LINKED ISSUE #943: Docker model runner container does not sync changes to config.json; requires restart (engine only)
### Description
Info: I am using Linux, Docker Engine-only, and no credential helper; I haven't tried any other use cases (I imagine a credential helper makes this trickier).

It appears that the docker-model-runner container takes its `config.json` from the local system on startup. However, this file is not synced from the system while the model-runner is running, meaning any `docker login` actions will not be reflected into the model runner, and any `docker model push` that requires the new auth will fail with a standard "insufficient scope". This is confusing as it suggests that some additional configuration on the registry side is required (regular pushes succeed).

Restarting the model-runner container syncs the updated config.json and credentials can be successfully resolved, so it's not a real issue, but there's no indication in the CLI that a restart is required or that credentials are not synced; I had to dig around to figure this out.

### Reproducing
```bash
# Stop runner and logout
docker model stop-runner
docker logout

# Start runner and login
docker model start-runner
docker login -u [username]

# Attempt to push a model
docker model push ... # fails, even though you just logged in

# Check config.json inside model runner
docker exec docker-model-runner cat /home/modelrunner/.docker/config.json  # does not contain new entry

# Restart model-runner and retry
docker model restart runner
docker model push ... # succeeds
docker exec docker-model-runner cat /home/modelrunner/.docker/config.json  # contains new entry
```

### Version

* docker model version:
    ```
    Client:
    Version:    v1.2.1
    OS/Arch:    linux/amd64

    Server:
    Version:    v1.2.1
    Engine:     Docker Engine
    ```

* docker version:
    ```
    Client: Docker Engine - Community
    Version:           29.5.2
    API version:       1.54
    Go version:        go1.26.3
    Git commit:        79eb04c
    Built:             Wed May 20 14:42:04 2026
    OS/Arch:           linux/amd64
    Context:           default

    Server: Docker Engine - Community
    Engine:
      Version:          29.5.2
      API version:      1.54 (minimum version 1.40)
      Go version:       go1.26.3
      Git commit:       568f755
      Built:            Wed May 20 14:38:16 2026
      OS/Arch:          linux/amd64
      Experimental:     false
    containerd:
      Version:          v2.2.4
      GitCommit:        193637f7ee8ae5f5aa5248f49e7baa3e6164966e
    runc:
      Version:          1.3.5
      GitCommit:        v1.3.5-0-g488fc13e
    docker-init:
      Version:          0.19.0
      GitCommit:        de40ad0
    ```

areebahmeddd:
## Summary

after `docker login` runs while the model runner container is already up, the container continues using a stale copy of `~/.docker/config.json` until restarted.

before each push, pull, or compose model download, copy the host's `~/.docker/config.json` into the container so the latest credentials are used. this only applies to docker engine (moby), since desktop and cloud manage credentials independently.

Fixes #943


cmd/cli/pkg/standalone/containers.go
```diff
@@ -28,10 +28,9 @@ import (
...
 
-// copyDockerConfigToContainer copies the Docker config file from the host to the container
-// and sets up proper ownership and permissions for the modelrunner user.
-// It does nothing for Desktop and Cloud engine kinds.
-func copyDockerConfigToContainer(ctx context.Context, dockerClient *client.Client, containerID string, engineKind types.ModelRunnerEngineKind) error {
+// SyncDockerConfigToContainer copies the host's ~/.docker/config.json into the
+// running container. It is a no-op for Desktop and Cloud engine kinds.
+func SyncDockerConfigToContainer(ctx context.Context, dockerClient *client.Client, containerID string, engineKind types.ModelRunnerEngineKind) error {
 	// Do nothing for Desktop and Cloud engine kinds
 	if engineKind == types.ModelRunnerEngineKindDesktop || engineKind == types.ModelRunnerEngineKindCloud ||
```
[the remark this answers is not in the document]
areebahmeddd: not applicable. this env var only affects desktop engines in test mode and does not change behavior for real moby engines

cmd/cli/commands/utils.go
```diff
@@ -247,6 +249,32 @@ func addRunnerFlags(cmd *cobra.Command, opts runnerFlagOptions) {
...
+// container. Only applicable for Moby engine setups; a no-op otherwise.
+func syncDockerConfigForRegistry(ctx context.Context, printer standalone.StatusPrinter) {
+	if modelRunner == nil {
+		return
+	}
+	engineKind := modelRunner.EngineKind()
+	if engineKind != types.ModelRunnerEngineKindMoby {
+		return
+	}
+	dockerClient, err := desktop.DockerClientForContext(dockerCLI, dockerCLI.CurrentContext())
```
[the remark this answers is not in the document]
areebahmeddd: not applicable. dockercli is always initialized before this code runs

cmd/cli/pkg/standalone/containers.go
```diff
@@ -28,10 +28,9 @@ import (
 // controllerContainerName is the name to use for the controller container.
 const controllerContainerName = "docker-model-runner"
 
-// copyDockerConfigToContainer copies the Docker config file from the host to the container
-// and sets up proper ownership and permissions for the modelrunner user.
-// It does nothing for Desktop and Cloud engine kinds.
-func copyDockerConfigToContainer(ctx context.Context, dockerClient *client.Client, containerID string, engineKind types.ModelRunnerEngineKind) error {
+// SyncDockerConfigToContainer copies the host's ~/.docker/config.json into the
+// running container. It is a no-op for Desktop and Cloud engine kinds.
+func SyncDockerConfigToContainer(ctx context.Context, dockerClient *client.Client, containerID string, engineKind types.ModelRunnerEngineKind) error {
```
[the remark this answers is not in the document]
areebahmeddd: not applicable. this function is only called with a valid client in normal flows

cmd/cli/commands/utils.go
```diff
@@ -247,6 +249,32 @@ func addRunnerFlags(cmd *cobra.Command, opts runnerFlagOptions) {
...
+	if err != nil {
+		return // non-fatal: proceed with potentially stale credentials
```
sathiraumesh: may be better to put a log message  of type warning if not the error is just missed

cmd/cli/commands/utils.go
```diff
@@ -247,6 +249,32 @@ func addRunnerFlags(cmd *cobra.Command, opts runnerFlagOptions) {
...
+	}
+	defer dockerClient.Close()
+
+	containerID, _, _, err := standalone.FindControllerContainer(ctx, dockerClient)
+	if err != nil || containerID == "" {
+		return // non-fatal: container may not be running yet
```
sathiraumesh: same here can add a log

cmd/cli/pkg/standalone/containers_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"context"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/docker/model-runner/cmd/cli/pkg/types"
+)
+
+// TestSyncDockerConfigToContainer_NoopForDesktopAndCloud verifies that
+// SyncDockerConfigToContainer skips Desktop and Cloud engine kinds.
```
sathiraumesh, line 10: verifies that SyncDockerConfigToContainer skips Desktop and Cloud engine kinds. can simplfy to

sathiraumesh (changes requested):

cmd/cli/pkg/standalone/containers.go
```diff
@@ -28,17 +29,16 @@ import (
...
```
sathiraumesh, line 33: I'm wondering why you changed the the name of the function any reason ?

cmd/cli/pkg/standalone/containers.go
```diff
@@ -28,17 +29,16 @@ import (
...
 	// Do nothing for Desktop and Cloud engine kinds
 	if engineKind == types.ModelRunnerEngineKindDesktop || engineKind == types.ModelRunnerEngineKindCloud ||
 		os.Getenv("_MODEL_RUNNER_TREAT_DESKTOP_AS_MOBY") == "1" {
 		return nil
 	}
 
-	dockerConfigPath := os.ExpandEnv("$HOME/.docker/config.json")
+	dockerConfigPath := filepath.Join(dockercliconfig.Dir(), "config.json")
```
sathiraumesh: I checked with the previous implementation and the current one they have different ways internally for example if someone hase set `DOCKER_CONFIG ` env it would break. May be this current change is the correct way need to check with the authors @ericcurtin @ilopezluna

cmd/cli/pkg/standalone/containers.go
areebahmeddd: um that's mostly ai's doing. it probably decided that name reflects the functionality better. (i still verify changes with make validate, build and review the code myself : )

cmd/cli/pkg/standalone/containers.go
ericcurtin: I suggest comparing against what moby/containerd does and copying that

areebahmeddd: i used config.Dir() from docker/cli here because that's the same logic docker uses to locate its config directory

for the moby/containerd point.. my understanding is that this part is handled on the  cli side rather than by the daemon itself

sathiraumesh: @areebahmeddd yes this might be the normal way but always make sure nothing breaks for backwards compatibility. if you check the implementations they differ a bit that's why i pointed it out.

areebahmeddd: could you point out for me? 🤔

sathiraumesh: so this is from the `cli package` config implementation what you are using now 
```
const (
	// EnvOverrideConfigDir is the name of the environment variable that can be
	// used to override the location of the client configuration files (~/.docker).
	//
	// It takes priority over the default, but can be overridden by the "--config"
	// command line option.
	EnvOverrideConfigDir = "DOCKER_CONFIG"

	// ConfigFileName is the name of the client configuration file inside the
	// config-directory.
	ConfigFileName = "config.json"
	configFileDir  = ".docker"
	contextsDir    = "contexts"
)
```
...
```
func Dir() string {
	initConfigDir.Do(func() {
		configDir = os.Getenv(EnvOverrideConfigDir)
		if configDir == "" {
			configDir = filepath.Join(getHomeDir(), configFileDir)
		}
	})
	return configDir
}
```

You see the implementation is different. thats what i was pointing to like it considers some extra env overrides may be the users are using now which might break

areebahmeddd: i see .. let me try something

areebahmeddd:
gentle ping! @ericcurtin

ericcurtin:
I'm on PTO, maybe @ilopezluna or @doringeman can look at this, otherwise I will take a look when I get back

cmd/cli/commands/utils.go
```diff
@@ -247,6 +249,37 @@ func addRunnerFlags(cmd *cobra.Command, opts runnerFlagOptions) {
 	}
 }
 
+// syncDockerConfigForRegistry copies the host's Docker config into the running
+// container. Only applicable for Moby engine setups; a no-op otherwise.
+func syncDockerConfigForRegistry(ctx context.Context, printer standalone.StatusPrinter) {
```
doringeman, line 254: This needs the `!vllmOnWSL` guard which you can find in containers.go:631.

doringeman (commented):
Thanks!
Added a note regarding a the `vllmOnWSL` guard.
Moreover, I think it'd be good to improve the security around this and stop copying config.json.

cmd/cli/commands/utils.go
areebahmeddd: added !

doringeman (approved):

PR #955, opened 2026-06-04
chore: bump llama.cpp to b9501

ilopezluna:


ericcurtin (approved):

ericcurtin (approved):

.versions
ericcurtin, line 1: We should do something about this file, manually editing this really isn't great

PR #965, opened 2026-06-10
chore: bump llama.cpp to b9587

ilopezluna:


ilopezluna:
Superseded by #967, which bumps to b9592 instead.

PR #968, opened 2026-06-11
ci(release): prune cloud builder cache before building

ilopezluna:
## Context

The release for the llama.cpp `b9501 → b9592` bump failed in the `build` job with:

```
ResourceExhausted: failed to compute cache key: symlink ... /buildkit/data/runc-overlayfs/snapshots/... : no space left on device
```

The error fires on the shared **Docker Build Cloud** builder (`driver: cloud`, endpoint `docker/make-product-smarter`) while unpacking an upstream llama.cpp image snapshot.

## Investigation

The upstream `ghcr.io/ggml-org/llama.cpp` images grew notably across this bump (compressed `linux/amd64` layers, measured against GHCR):

| Variant (tag base)    | b9501   | b9592   | Δ            |
|-----------------------|---------|---------|--------------|
| `server-vulkan` (cpu) | 160 MB  | 296 MB  | **+85%**     |
| `server-openvino`     | 308 MB  | 540 MB  | **+75%**     |
| `server-musa`         | 737 MB  | 921 MB  | +25%         |
| `server-cuda13`       | 1543 MB | 1800 MB | +17%         |
| `server-rocm`         | 7183 MB | 7322 MB | +2%          |

**Root cause:** the release builds 7 variants (cpu/cuda on amd64+arm64) on one shared cloud builder with no cache cleanup. Cache accumulated across runs filled the disk; the ~1 GB of compressed image growth (several GB uncompressed across all variants × platforms) was the final push.

## Change

Add a `docker buildx prune -af` step right after *Set up Buildx* so every release starts with a clean builder disk.

> Note: this prevents future failures. To unblock the **current** stuck release, the cloud builder cache must be purged manually (`docker buildx prune -af --builder cloud-docker-make-product-smarter`) and the run re-run.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ericcurtin (approved):

ericcurtin:
Approved @ilopezluna ... Sorcery has the right idea I think updates should be in one place, for another PR merging...

ilopezluna:
@ericcurtin These ARGs are only defaults for local/dev builds, in the release they're always overridden by build-args derived from `.versions` via `scripts/resolve-llama-upstream-image.sh`, which is the actual single source of truth. The suggestion also only half-solves it: `server-vulkan` (the variant tag base) stays hardcoded, so the image still can't 
be derived from the version alone. Not a real skew risk here.

PR #970, opened 2026-06-13
fix: add Pascal CUDA architectures

### LINKED ISSUE #929: Add Pascal GPU (sm_61/sm_62) support to llama.cpp CUDA build
## Summary

Docker Model Runner's bundled llama.cpp binary is compiled for CUDA architectures
`750, 800, 860, 890, 1200, 1210` (Turing and newer), which excludes Pascal GPUs
(GTX 1080, 1080 Ti, 1070, 1060, etc. — `sm_61/sm_62`).

As a result, users with Pascal GPUs see the following at startup:

```
ggml_cuda_init: failed to initialize CUDA: forward compatibility was attempted on non supported HW
warning: no usable GPU found, --gpu-layers option will be ignored
```

And the runner logs `gpuSupport=false`, falling back entirely to CPU inference.

## Environment

- GPU: NVIDIA GeForce GTX 1080 Ti (Pascal, sm_61)
- Driver: 580.159.03 / CUDA 13.0
- Docker Model Runner: v1.1.38 (Docker Engine, Linux)
- llama.cpp backend: Running (fc2b0053f)

## Expected behavior

llama.cpp should use the GPU for inference on Pascal hardware, as CUDA 12/13
still fully supports `sm_60`, `sm_61`, and `sm_62` via nvcc.

## Root cause

The [`llamacpp/native/cuda.Dockerfile`](https://github.com/docker/model-runner/blob/main/llamacpp/native/cuda.Dockerfile)
does not set `-DCMAKE_CUDA_ARCHITECTURES`, so CMake uses the CUDA SDK defaults
which no longer include Pascal in recent versions.

## Suggested fix

Add Pascal architectures to the CMake flags in `cuda.Dockerfile`:

```dockerfile
RUN echo "-B build \
    -DGGML_CUDA=ON \
    -DCMAKE_CUDA_ARCHITECTURES='61;62;70;75;80;86;89;120;121' \
    ..." > cmake-flags
```

This would extend GPU support to a large installed base of Pascal GPUs with no
functional downside — only a slightly larger binary and longer build time.

nanookclaw:
## Summary

Closes #929.

This sets `CMAKE_CUDA_ARCHITECTURES` explicitly for the native llama.cpp CUDA build so the image includes Pascal targets `sm_61` and `sm_62`. The list follows the issue's suggested architecture set and keeps the existing newer CUDA targets covered:

`61;62;70;75;80;86;89;120;121`

Without an explicit value, recent CUDA/CMake defaults can omit Pascal, which leaves GTX 10-series and similar cards without a compatible CUDA kernel and causes Docker Model Runner to fall back to CPU inference.

## Validation

- `git diff --check`
- Confirmed the Dockerfile contains the architecture flag exactly once
- Confirmed the semicolon-separated CMake list remains a single argument when passed through the existing `cmake-flags` command-substitution pattern

I did not run a full CUDA image build locally because that requires pulling the NVIDIA CUDA builder image and compiling llama.cpp; there does not appear to be a focused lightweight Dockerfile flag test in this repo.

### DIFF
--- a/llamacpp/native/cuda.Dockerfile
+++ b/llamacpp/native/cuda.Dockerfile
@@ -32,6 +32,7 @@ RUN echo "-B build \
     -DGGML_NATIVE=OFF \
     -DGGML_OPENMP=OFF \
     -DGGML_CUDA=ON \
+    -DCMAKE_CUDA_ARCHITECTURES=61;62;70;75;80;86;89;120;121 \
     -DCMAKE_CUDA_COMPILER=/usr/local/cuda/bin/nvcc \
     -DLLAMA_OPENSSL=OFF \
     -GNinja \


ericcurtin:
Eventually we should use to using llama.cpp artefacts from llama.cpp directly... We will support whatever they deem suitable to support...

@ilopezluna PTAL

ilopezluna:
Thanks a lot for this, @nanookclaw, and sorry for the duplicate effort!

Closing this in favor of #986, which already landed the equivalent CMAKE_CUDA_ARCHITECTURES flag (61;62;70;75;80;86;89) in main.

One important heads-up for anyone finding this later: on Linux, Docker Model Runner no longer builds llama.cpp from llamacpp/native/cuda.Dockerfile. The Linux runtime ships the upstream ggml-org image directly (Dockerfile → FROM ${LLAMA_UPSTREAM_IMAGE} → ghcr.io/ggml-org/llama.cpp:server-cuda13-*, resolved in scripts/resolve-llama-upstream-image.sh), and download_linux.go has runtime backend updates disabled. cuda.Dockerfile only feeds the docker/docker-model-backend-llamacpp image consumed by Docker Desktop on macOS/Windows.

So this change (and #986) won't affect Linux users like the reporter in #929 (GTX 1080 Ti on Docker Engine). Pascal support on Linux needs to come from the upstream server-cuda13 build including sm_61/sm_62. 
Thanks again! 🙏

PR #1010, opened 2026-07-13
Add structured output support to Responses API

### LINKED ISSUE #905: Full support for v1/responses endpoint.
In the open AI doc, the `completions` endpoint is considered legacy. I noticed there was partial support, but it would be nice to have full support including structured output like this:

` text: {
                    format: {
                        type: "json_schema",
                        name: "answer_prompt",
                        strict: true,
                        schema: schema
                    }
                }`

dmedovich:
Fixes #905

Adds support for Responses API structured output by mapping `text.format`
to the existing Chat Completions `response_format` request sent to the backend.

The issue specifically calls out `/engines/v1/responses` and structured output
handling where the backend supports it.

## Approach

Responses requests use:

```json
{
  "text": {
    "format": {
      "type": "json_schema",
      "name": "answer_prompt",
      "strict": true,
      "schema": {}
    }
  }
}
```

The existing backend path expects the Chat Completions shape instead, so this PR
translates that into:

```json
{
  "response_format": {
    "type": "json_schema",
    "json_schema": {
      "name": "answer_prompt",
      "strict": true,
      "schema": {}
    }
  }
}
```

`json_object` is mapped the same way to `response_format.type = json_object`.
The original `text` config is kept on the returned Response object.

Also added `/engines/v1/responses` and `/engines/{backend}/v1/responses`, matching
the existing engine-qualified Chat Completions routes.

## Behavior

`store: false` is now honored for both normal and streaming responses: the
response is returned, but not saved for later lookup.

Unsupported Responses-only fields now return `400` instead of being silently
ignored. `background: false`, `truncation: "disabled"`, and explicit `null`
hosted fields are treated as no-ops.

## Trade-offs to flag

This stays on top of the existing Responses compatibility layer. It does not
replace streaming with native OpenAI Responses typed events.

I'm happy to work on that larger native Responses API follow-up too if we want
to take it in that direction.

## tested

- ✅ `git diff --check`
- ✅ `go test ./pkg/responses ./pkg/routing`
- ✅ `go test -race ./pkg/responses ./pkg/routing ./pkg/inference/scheduling`
- ✅ `make validate-all`
- ✅ `make e2e`

`make e2e` includes a new structured output case through `/responses`.

### DIFF
--- a/e2e/inference_test.go
+++ b/e2e/inference_test.go
@@ -120,7 +120,10 @@ func TestE2E_Inference(t *testing.T) {
 								"schema": map[string]any{
 									"type": "object",
 									"properties": map[string]any{
-										"answer": map[string]any{"type": "string"},
+										"answer": map[string]any{
+											"type": "string",
+											"enum": []string{"yes"},
+										},
 									},
 									"required":             []string{"answer"},
 									"additionalProperties": false,
@@ -150,6 +153,9 @@ func TestE2E_Inference(t *testing.T) {
 				}
 				if output.Answer == "" {
 					t.Fatalf("answer field is empty in output_text=%q", resp.OutputText)
+				}
+				if output.Answer != "yes" {
+					t.Fatalf("answer = %q, want yes (output_text=%q)", output.Answer, resp.OutputText)
 				}
 			})
 
--- a/e2e/inference_test.go
+++ b/e2e/inference_test.go
@@ -106,6 +106,53 @@ func TestE2E_Inference(t *testing.T) {
 				t.Logf("responses API: %q", resp.OutputText)
 			})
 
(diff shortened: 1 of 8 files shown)


dmedovich:
[sourcery-ai](https://github.com/apps/sourcery-ai)
with the current nested `ServeMux` setup, `routing.NewRouter` still needs to decide which top-level paths are delegated to the Responses handler, while `responses.NewHTTPHandler` needs to register the concrete method/path patterns on its own mux. So I kept the existing boundary and avoided a larger router refactor in this PR

ericcurtin:
Build failed, needs a fix

dmedovich:
@ericcurtin  I'll take a look a little later

ericcurtin (approved):

PR #1037, opened 2026-08-11
fix(cli): remove unused backend flag from compose up

### LINKED ISSUE #1036: compose up: --backend flag is accepted but never applied
﻿## Bug

`docker model compose up` defines a `--backend` flag (default `llama.cpp`) but never uses the value when calling `ConfigureBackend`.

## Evidence (main)

In `cmd/cli/commands/compose.go` (`newUpCommand`):

```go
var backend string
// ...
c.Flags().StringVar(&backend, "backend", llamacpp.Name, "inference backend to use")
```

The only write is the flag binding. `backend` is never read. Configure is built as:

```go
desktopClient.ConfigureBackend(scheduling.ConfigureRequest{
    Model:                model,
    BackendConfiguration: backendConfig,
    RawRuntimeFlags:      rawRuntimeFlags,
})
```

There is no backend name field set from the flag. `scheduling.ConfigureRequest` also has no top-level backend selector (only model + `BackendConfiguration` + runtime flags). So even if the flag were plumbed, the request shape would need a place to put it (for example how `ModelConfigEntry.Backend` is stored elsewhere).

Meanwhile `compose metadata` advertises `--backend` via `commandParameters(upCmd)`, so Compose/UI consumers may show a backend selector that does nothing.

## Expected

Either:

1. Plumb `--backend` into the configure/install path so non-default backends work under compose, or
2. Remove the flag from `up` and from metadata until the API can honor it (so metadata matches behaviour).

## Impact

Users (or Compose integrations) that pass `--backend vllm` (or anything other than the default) get silent no-ops: the flag is accepted, metadata lists it, but configuration always follows the runner default.

## Related

- Compose only otherwise configures context-size / runtime-flags / speculative fields (#1035 covers env separately).

I can send a PR either wiring backend selection or dropping the dead flag after maintainers pick a direction.

VedantMadane:
### Summary

Fixes #1036

In cmd/cli/commands/compose.go, 
ewUpCommand() defined a --backend flag variable ackend but never passed it to desktopClient.ConfigureBackend(...). As a result, passing --backend was a silent no-op while compose metadata still advertised --backend parameter to UI consumers.

This PR removes the dead --backend flag variable and flag binding from 
ewUpCommand() and removes the unused llamacpp import.

### DIFF
--- a/cmd/cli/commands/compose.go
+++ b/cmd/cli/commands/compose.go
@@ -12,7 +12,6 @@ import (
 	"github.com/docker/model-runner/cmd/cli/desktop"
 	"github.com/docker/model-runner/cmd/cli/pkg/types"
 	"github.com/docker/model-runner/pkg/inference"
-	"github.com/docker/model-runner/pkg/inference/backends/llamacpp"
 	dmrm "github.com/docker/model-runner/pkg/inference/models"
 	"github.com/docker/model-runner/pkg/inference/scheduling"
 	"github.com/spf13/cobra"
@@ -37,7 +36,6 @@ func newUpCommand() *cobra.Command {
 	var models []string
 	var ctxSize int64
 	var rawRuntimeFlags string
-	var backend string
 	var draftModel string
 	var numTokens int
 	var minAcceptanceRate float64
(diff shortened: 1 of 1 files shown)


doringeman:
Hello @VedantMadane,

I need to look into why this is here. Indeed `backend` is not present in https://github.com/docker/compose/blob/main/pkg/compose/model.go.

